PR #3, opened 2024-01-22
Fixup CI

asmeurer:


asmeurer:
And I guess I have to merge this for the other workflow to run.

asmeurer:
What matrix of numpy/python versions should we support here? Everything in https://scientific-python.org/specs/spec-0000/? Or is only supporting numpy 1.26 and 2.0 enough?

rgommers:
Only 1.26 and 2.0 should do I believe.

PR #9, opened 2024-01-24
Update documentation

asmeurer:


asmeurer:
@rgommers if you have a chance, can you review this?

In addition to the changes in this PR, can you also take a look at the repo as a whole. I think I have everything ready to do an initial release, but I don't know if I forgot anything. 

I also opened some issues on this repo for ideas of things we can do in the future. But I want to keep the first release as close to the current numpy.array_api as possible.

CHANGELOG.md
```diff
@@ -0,0 +1,30 @@
+# array-api-strict Changelog
+
+## 1.0 (????)
```
rgommers, line 3: 1.0 SGTM

CHANGELOG.md
```diff
@@ -0,0 +1,30 @@
...
+
+This is the first release of `array_api_strict`. It is extracted from
+`numpy.array_api`, which was included as an experimental submodule in NumPy
+versions prior to 2.0. Note that the commit history in this repository is
+extracted from the git history of numpy/array_api/ (see the [README](README.md)).
+
+Additionally, the following changes are new to `array_api_strict` from
+`numpy.array_api` in NumPy 1.26 (the last major NumPy release to include
```
rgommers, line 11:
```suggestion
`numpy.array_api` in NumPy 1.26 (the last NumPy feature release to include
```

README.md
```diff
@@ -1,13 +1,191 @@
...
+
+- Only those functions and methods that are [defined in the
+  standard](https://data-apis.org/array-api/latest/API_specification/index.html)
+  are included.
+
+- In those functions, only the keyword-arguments that are defined by the
+  standard are included. All signatures in array-api-strict use
+  [positional-only
+  arguments](https://data-apis.org/array-api/latest/API_specification/function_and_method_signatures.html#function-and-method-signatures).
+  As noted above, only array_api_strict array objects are accepted by
```
rgommers, line 91: formatting: `array_api_strict` needs backticks

README.md
```diff
@@ -1,13 +1,191 @@
...
+  functions, except in the places where the standard allows Python scalars
+  (i.e., functions to not automatically call `asarray` on their inputs).
```
rgommers, line 93: this sentence reads slightly awkward to me. maybe it's a typo (to not -> do not)?

README.md
```diff
@@ -1,13 +1,191 @@
...
+  the input dtype(s) to be supported. This is one of the most restrictive
+  aspects of the library. For example, in NumPy, most transcendental functions
+  like `sin` will accept integer array inputs, but the [standard only requires
+  them to accept floating-point
+  inputs](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sin.html#array_api.sin),
+  so in array-api-strict, `sin(integer_array)` will raise an exception.
+
+- The
+  [indexing](https://data-apis.org/array-api/latest/API_specification/indexing.html)
+  semantics required by the standard are not
```
rgommers, line 109: incomplete sentence here

README.md
```diff
@@ -1,13 +1,191 @@
...
+  arrays.
+
+- Dtype objects are just empty objects that only implement [equality
+  comparison](https://data-apis.org/array-api/latest/API_specification/generated/array_api.data_types.__eq__.html).
+  The way to access dtype objects in the standard is by name, like
+  `xp.float32`.
+
+- The array object type itself is private and should not be accessed.
+  Subclassing or otherwise trying to directly initialize this object is not
+  supported. Arrays should created with one of the [array creation
```
rgommers, line 121: typo: "should be created"

README.md
```diff
@@ -1,13 +1,191 @@
...
+  functions](https://data-apis.org/array-api/latest/API_specification/creation_functions.html)
+  such as `asarray`.
+
+## Caveats
+
+array-api-strict is a thin pure Python wrapper around NumPy. NumPy 2.0 fully
```
rgommers, line 127: I suggest: "NumPy 2.0 has nearly full support for ....", since there are a few aspects that don't comply (e.g., returning scalars)

array_api_strict/__init__.py
```diff
@@ -1,117 +1,18 @@
...
-- Some functions are not yet fully tested in the array API test suite, and may
-  require updates that are not yet known until the tests are written.
-
-- The spec is still in an RFC phase and may still have minor updates, which
-  will need to be reflected here.
-
-- Complex number support in array API spec is planned but not yet finalized,
-  as are the fft extension and certain linear algebra functions such as eig
-  that require complex dtypes.
+array_api_strict is a strict, minimal implementation of the Python array
```
rgommers, line 2: need backticks here as well

array_api_strict/__init__.py
```diff
@@ -1,117 +1,18 @@
...
+API (https://data-apis.org/array-api/latest/)
```
rgommers, line 3: this probably should be a markdown link around "Python array API"? and append "standard"?

rgommers (approved):
This looks quite good Aaron, I only have a few minor comments.

rgommers:
> In addition to the changes in this PR, can you also take a look at the repo as a whole. I think I have everything ready to do an initial release, but I don't know if I forgot anything.

I think it can be shipped as is, however you may want to do some of the following (can be done later):

- Packaging: we may want to be good citizens and use `pyproject.toml` here, and get rid of all `setup.*` content
- `versioneer` is pretty awful, and this package is really simple. I'd just use `1.0` and then `1.1-dev` in `pyproject.toml` and leave it at that, rather than use thousands of lines of questionable-quality code to support a simple version string.

> I also opened some issues on this repo for ideas of things we can do in the future. But I want to keep the first release as close to the current numpy.array_api as possible.

Good ideas, +1 to both the first release plan and all of those things as follow-ups.

rgommers:
Looks like the switchover is easy and everything works as expected for SciPy.

array_api_strict/__init__.py
asmeurer: It's not a markdown link here because this is the docstring in a Python file.

README.md
asmeurer: I'll change this, but I was under the impression that scalars were not actually breaking compliance. As long as scalars duck-type as 0-D arrays, which they do, then it's just an implementation detail that numpy has a separate type for them.

asmeurer:
I've used versioneer on dozens of projects and never had any issues with it. All the code in versioneer is used for generating development version numbers and doing packaging. The actual resulting package has a very simple file generated in it. Here's what it creates for this branch

```py
# This file was generated by 'versioneer.py' (0.29) from
# revision-control system data, or from the parent directory name of an
# unpacked source archive. Distribution tarballs contain a pre-generated copy
# of this file.

import json

version_json = '''
{
 "date": "2024-01-23T17:55:23-0700",
 "dirty": false,
 "error": null,
 "full-revisionid": "c2d570061e8434b58c4e6609e2a849cf75e6a390",
 "version": "0+untagged.193.gc2d5700"
}
'''  # END VERSION_JSON


def get_versions():
    return json.loads(version_json)
```

Admittedly, even that is more complicated than it could be. It could just generate a file with just `__version__ = '<the version number>'. But it's way simpler for *runtime* than, e.g., setuptools_scm, which [grabs the version number from the package metadata](https://setuptools-scm.readthedocs.io/en/latest/usage/#version-at-runtime). I don't like that at all. If you know of a simpler tool I'm open to using it, but I do like to have something that lets me write the version number just once (in the tag). Otherwise I forget to update it, like I [recently did for array-api-compat](https://github.com/data-apis/array-api-compat/commit/05548f0281f13e7cb6a293a422ef8c1a90021413) (we can't use versioneer there because of the vendorability requirement). I don't actually care as much about the git hash version thing, although it is nice to have and it occasionally helps with debugging user issues.

README.md
rgommers: maybe, not sure if the duck-typing really is 100%. there's a few other reasons though, see the "not adopted part in the PR for NEP 56

rgommers:
> If you know of a simpler tool I'm open to using it, but I do like to have something that lets me write the version number just once (in the tag).

Imho none of these tools are worth it, for a one-line change per release. `versioneer` specifically we've had problems with in NumPy, and since dropped - as did Matplotlib and others. If you've had to debug `versioneer` code just once, you notice it's 2000+ lines of bad code _generating_ 500+ lines more code to then ship.

But okay, I don't have to make the updates here, so your call.

README.md
asmeurer: Where is NEP 56? Google comes up with nothing.

asmeurer:
OK, I'm going to try to do a release. Based on my experience with array-api-compat, getting the auto-deploy to work the first time around can be a little annoying, so I may end up pushing a `1.0` tag to this repo multiple times.

asmeurer:
OK, array-api-strict 1.0 has been released. Now I need to create a conda-forge package.

README.md
rgommers: https://github.com/numpy/numpy/pull/25542

rgommers:
Great, thanks!

README.md
asmeurer: I'm going to continue this discussion over there since this is closed, but that NEP seems to just state that scalars aren't in compliance without actually explaining why. https://github.com/numpy/numpy/pull/25542#issuecomment-1909047579

Issue #11, opened 2024-03-01
MAINT: copy semantics shim


tylerjereddy:
Probably not surprising; in https://github.com/scipy/scipy/pull/20172 I noticed what looks like a requirement for at least a small `copy` argument pass-through shim with NumPy `main` here because of (currently) a single test in SciPy. In that PR I was able to patch similar errors with fairly crude `copy=None` pass-through type shims, though in this library it may actually need to do something useful (the intended purpose) if the `copy` arg really needs to be respected I suppose.

```
============================================================================================ FAILURES =============================================================================================
______________________________________________________________________________ test_dispatch_to_unrecognize_library _______________________________________________________________________________
[gw22] linux -- Python 3.11.2 /home/treddy/python_venvs/py_311_scipy_dev/bin/python
scipy/special/tests/test_support_alternative_backends.py:25: in test_dispatch_to_unrecognize_library
    res = f(xp.asarray(x))
        f          = <function get_array_special_func.<locals>.f at 0x7f404756a2a0>
        x          = [1, 2, 3]
        xp         = <module 'array_api_strict' from '/home/treddy/python_venvs/py_311_scipy_dev/lib/python3.11/site-packages/array_api_strict/__init__.py'>
scipy/special/_support_alternative_backends.py:33: in f
    array_args = [np.asarray(arg) for arg in array_args]
        args       = (Array([1, 2, 3], dtype=array_api_strict.int64),)
        array_args = (Array([1, 2, 3], dtype=array_api_strict.int64),)
        f_scipy    = <ufunc 'ndtr'>
        kwargs     = {}
        n_array_args = 1
        other_args = ()
        xp         = <module 'array_api_strict' from '/home/treddy/python_venvs/py_311_scipy_dev/lib/python3.11/site-packages/array_api_strict/__init__.py'>
scipy/special/_support_alternative_backends.py:33: in <listcomp>
    array_args = [np.asarray(arg) for arg in array_args]
E   TypeError: Array.__array__() got an unexpected keyword argument 'copy'
        .0         = <tuple_iterator object at 0x7f4062090cd0>
        arg        = Array([1, 2, 3], dtype=array_api_strict.int64)
===================================================================================== short test summary info =====================================================================================
FAILED scipy/special/tests/test_support_alternative_backends.py::test_dispatch_to_unrecognize_library - TypeError: Array.__array__() got an unexpected keyword argument 'copy'

```

asmeurer:
[...]

PR #12, opened 2024-03-02
Support `copy` keyword in `__array__` method

rgommers:
Closes gh-11


rgommers:
The one failure is unrelated (that's the `test_solve` one), but it cancelled a bunch of test jobs.

asmeurer:
Tests should be fixed by https://github.com/data-apis/array-api-strict/pull/14

rgommers:
> Is this a change that any library that implements `__array__` will have to make?

Yes indeed - missing `copy` or `dtype` keywords from the `__array__` signature will emit a `UserWarning` in 2.0, and will probably be a deprecation later.

PR #15, opened 2024-03-05
Fix asarray(copy=False)

asmeurer:
For NumPy 2.0, this is implemented directly. For NumPy 1, we emulate it by checking if asarray() creates a copy or not.

This also removes support for the np._CopyMode enum in asarray(), as this is not portable.

@rgommers I can't remember if there was a good reason that we didn't just emulate copy=False in this way originally (instead, we raised `NotImplementedError`).

### DIFF
--- a/array-api-tests-xfails.txt
+++ b/array-api-tests-xfails.txt
@@ -1,7 +1,3 @@
-# copy=False is not yet implemented
-# https://github.com/numpy/numpy/pull/25168
-array_api_tests/test_creation_functions.py::test_asarray_arrays
-
 # Some fft tests are currently wrong
 # (https://github.com/data-apis/array-api-tests/issues/231)
 array_api_tests/test_fft.py::test_fft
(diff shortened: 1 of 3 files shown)


asmeurer:
Maybe it had to do with when `obj` is not an array. Are there instances where copy=False could fail with the buffer protocol?

rgommers:
> Maybe it had to do with when `obj` is not an array. Are there instances where copy=False could fail with the buffer protocol?

No, neither the buffer protocol nor `__array_interface__` have that.

asmeurer:
Searched through the discussion at https://github.com/numpy/numpy/pull/18585 and couldn't find anything. I think I just wasn't really clear at the time that numpy copy=False worked like copy=None.

asmeurer:
OK, I updated this to handle Python built-ins and buffer protocol objects properly, and added some more tests since I don't think the test suite really checks `copy` that sufficiently.

PR #20, opened 2024-03-12
Add a workflow to automatically merge dependabot updates

asmeurer:
Dependabot only updates GitHub Actions workflow versions, so there should be little risk in this. It shouldn't actually merge things unless CI passes because of required checks.


asmeurer:
No way to know if this works until the next dependabot PR.

PR #24, opened 2024-03-27
Update api_version argument to __array_namespace__

asmeurer:
- Allow 2022.12
- Warn on 2021.12
- Be more strict about only accepting those two versions

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -17,6 +17,8 @@ from __future__ import annotations
 
 import operator
 from enum import IntEnum
+import warnings
+
 from ._creation_functions import asarray
 from ._dtypes import (
     _DType,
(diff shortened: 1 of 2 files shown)


ogrisel:
I am not clear about the current version of the spec implemented by `array-api-strict`, in particular the README does not seem to have any info about this.

I could not find some of the recently introduced function such as `xp.searchsorted` so this confirms that it does not yet implement the full `v2023.12` it would be great to have a table somewhere that map `array-api-strict` versions with Array API spec versions and maybe links to one or more tracking issues with the list of things to add/or change before the next versions can of the spec can be considered as supported.

asmeurer:
Only 2022.12 is currently implemented. You're right that I do need to open a tracking issue for 2023.12 support.

PR #28, opened 2024-03-29
Add basic documentation, based on the README

### LINKED ISSUE #27: Real documentation
Real documentation will be useful for array-api-strict, especially as we start to implement more features like #8. We can just copy the docs skeleton from array-api-compat.

asmeurer:
The documentation configuration is based on the configuration from
array-api-compat.

Fixes #27


asmeurer:
There isn't really any new documentation I intend to write here, but there will be some in the future as we build out more features.

asmeurer:
https://data-apis.org/array-api-strict/

PR #30, opened 2024-04-08
array-api-strict flags

### LINKED ISSUE #7: Add a mode that disables data-dependent shapes
Currently we support all the behaviors that require [data dependent shapes](https://data-apis.org/array-api/draft/design_topics/data_dependent_output_shapes.html) (currently just the `unique_*` functions, `nonzero`, and boolean indexing). 

It would be useful to have some sort of mode to disable these so people could test against libraries that don't have them.

### LINKED ISSUE #8: Support for older versions of the standard
Should we support older versions of the standard?

I don't think there's any point to supporting 2021.12, since no implementation actually supports it (including array-api-compat). However, as libraries actually start implementing the standard, being able to support multiple versions here could be useful. 

For now, we could just tell people to install older versions of array-api-strict. But if people actually test against old versions of the standard, I would much rather just support them at once here than trying to support backport versions. 

Plus, the `__array_namespace__` method has the `version` flag specifically for this sort of thing. 

By the way, if you were to ask the same question for array-api-compat, or any actual implementation for that manner, I would say no, unless the standard ends up adopting some serious backwards incompatible changes in some future version. But there could be an argument for the strict implementation to have this.

asmeurer:
Flags are global variables that set array-api-strict in a specific mode. Currently support flags change the support array API standard version, enable or disable data-dependent shapes, and enable or disable optional extensions.

So far I have only set up the structure for setting and getting these flags. Let me know if this seems like a good API layout or not.

I plan to add functionality to the various functions to enable/disable behaviors depending on what flags are set. 

Fixes #8 
Fixes #7

### DIFF
--- a/array_api_strict/_flags.py
+++ b/array_api_strict/_flags.py
@@ -11,31 +11,34 @@ Nothing in this file is part of the standard itself. A typical array API
 library will only support one particular configuration of these flags.
 """
 
+import functools
 import os
 
-supported_versions = [
+supported_versions = (
     "2021.12",
     "2022.12",
-]
+)
 
-STANDARD_VERSION = "2022.12"
+STANDARD_VERSION = default_version = "2022.12"
 
 DATA_DEPENDENT_SHAPES = True
 
-all_extensions = [
+all_extensions = (
     "linalg",
     "fft",
-]
+)
 
 extension_versions = {
     "linalg": "2021.12",
     "fft": "2022.12",
 }
 
-ENABLED_EXTENSIONS = [
+ENABLED_EXTENSIONS = default_extensions = (
     "linalg",
     "fft",
-]
+)
+
+# Public functions
 
 def set_array_api_strict_flags(
     *,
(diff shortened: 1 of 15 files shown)


array_api_strict/_flags.py
```diff
@@ -0,0 +1,289 @@
...
+      versions are: ``{supported_versions}``. The default version number is
+      ``{default_version!r}``.
+
+    - `data_dependent_shapes`: Whether data-dependent shapes are enabled in
+      array-api-strict. This flag is enabled by default. Array libraries that
+      use computation graphs may not be able to support functions whose output
+      shapes depend on the input data.
+
+      This flag is enabled by default. Array libraries that use computation graphs may not be able to support
+      functions whose output shapes depend on the input data.
```
rgommers, line 70: stray sentence, duplicate with paragraph above

rgommers:
@asmeurer I think the `get_`/`reset_`/`set_array_api_strict_flags` API is easy to understand, seems perfectly fine to me.

asmeurer:
This is ready to go. I'm not going to implement anything for 2021.12 unless someone requests it. But using this, I can add preliminary support for 2023.12. 

One thing I'm not completely sure about is if someone sets a flag using an environment variable, like `ARRAY_API_STRICT_DEPENDENT_SHAPES=False`, and then calls `reset_array_api_strict_flags()`, should that reset the flags back to what was set by the environment variables, or back to the default values? Right now I have it so that they are reset back to the default values.

PR #35, opened 2024-04-19
Preliminary 2023.12 support

asmeurer:
This will not be enabled by default by can be enabled with `set_array_api_strict_flags(api_version='2023.12')` or using `__array_namespace__(api_version='2023.12')`. It will issue a warning that not all functions are implemented and it hasn't been tested. 

My plan here is to just wrap the functions that are already in NumPy, and maybe add some of the straightforward inspection APIs. There's no test suite support yet, so any features missing from NumPy will probably need to wait. 

This issue has a list of the new APIs that need to be added for 2023.12. https://github.com/data-apis/array-api-tests/issues/249


asmeurer:
@honno the tests are failing here because I have this behavior:

```
>>> xp.signbit(xp.asarray([0., 1.]))
  File "/Users/aaronmeurer/Documents/array-api-strict/array_api_strict/_flags.py", line 293, in wrapper
    raise RuntimeError(
RuntimeError: The function signbit requires API version 2023.12 or later, but the current API version for array-api-strict is 2022.12
```

It can be enabled by setting the version flag to 2023.12, but I'd like to be able to test all supported versions here.

But the test suite uses `hasattr(xp, 'signbit')` and assumes that if it has `signbit` then it can use it. Can we append a simple sanity check to this test? Trying to make disabled functions actually raise AttributeError is a lot more complicated and I'd prefer not to do it if it isn't necessary.

asmeurer:
This is ready for review. However, we are blocked on the test suite issue I noted above. 

The only new API I'm missing is the new `device` and `copy` keywords to `from_dlpack`. Is it possible for me to implement those here without upstream NumPy support?

asmeurer:
Test suite should be fixed by https://github.com/data-apis/array-api-tests/pull/257. But I also need to add test runs for 2023.12. Apparently the signature and special cases are already running against 2023.12, so we can go ahead and add that.

asmeurer:
A lot of this failure on CI, which I've never seen before:

```
FAILED array_api_tests/test_searching_functions.py::test_argmin - hypothesis.errors.FailedHealthCheck: Examples routinely exceeded the max allowable size. (20 examples overran while generating 9 valid ones). Generating examples this large will usually lead to bad results. You could try setting max_size parameters on your collections and turning max_leaves down on recursive() calls.
See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.data_too_large to the suppress_health_check settings for this test.
```

I can't reproduce it locally. I guess we should just disable that health check.

asmeurer:
I opened https://github.com/data-apis/array-api-tests/issues/277 about the data_too_large health check issue. For now, I've just disabled that health check in the test suite so this should pass now.

PR #53, opened 2024-07-10
Allow any combination of real dtypes in comparisons

asmeurer:
This does not change == or != because the standard is currently unclear about that so I'd like to see what happens there first.

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -152,7 +152,13 @@ class Array:
     # spec in places where it either deviates from or is more strict than
     # NumPy behavior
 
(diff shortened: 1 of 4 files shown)


asmeurer:
Actually I went ahead and updated equal/not_equal/==/!= because it was easier.

asmeurer:
The test suite needs to be updated to test for this (and that will need to wait for update to the spec). I need this to unblock the CI here for #49 which is also blocking the test suite CI.

asmeurer:
It's looking like the standard is going to be updated to match the existing behavior, so this should be reverted https://github.com/data-apis/array-api/pull/822.

PR #59, opened 2024-09-04
Add "multi device" support

### LINKED ISSUE #56: Add virtual devices to make it easier for array API consumer to check that they use device correctly
Motivation: in scikit-learn, we run array API compliance tests both with `array_api_strict` and `array_api_compat.torch`. For the latter we run tests both with `cuda`, `mps` and `cpu` devices.

Testing with torch is important because it helps us reveal occurrences of problems related to device handling. For example let's consider the following problematic function:

```python
def sublinear_mask(data):
    return data <= xp.linspace(data[0], data[-1], num=data.shape[0])
```

Calling this with:

```python
sublinear_mask(xp.asarray([0, 1, 2, 2, 5, 5], device="mps"))
```

raises a `RuntimeError` because PyTorch does not implicitly move data across devices. Hence the bug should be fixed by changing the function to:

```python
def sublinear_mask(data):
    return data <= xp.linspace(data[0], data[-1], num=data.shape[0], device=data.device)
```

However, not all scikit-learn contributors have access to a machine with a non-cpu device (e.g. "mps" or "cuda") and therefore they have no easy way to detect this family of bugs by running their tests on their local laptop and only discover issues on the CI and need to use tools such as google colab to debug and refine their code instead of their regular dev environment.

To reduce this friction, it would be nice if `array-api-strict` could accept to create arrays with `a = xp.asarray([1, 2, 3], device="virtual_device_a")` and `b = xp.asarray([1, 2, 3], device="virtual_device_b")`  and would raise `RuntimeError` on operations that combine arrays from different devices as pytorch does.

betatim:
Having more than one device is useful during testing to allow you to find bugs related to how arrays on different devices are handled. Closes #56 

With scikit-learn we run into the frustrating situation were contributors execute tests locally, they all pass but then see failures on the CI related to the fact that e.g. PyTorch has several devices and some things work on the CPU device but not on the CUDA/MPS device. However, if you have neither of those on your local machine you can't really test this upfront and to debug it you need to rely on the CI.

The idea of this PR is to add support for multiple devices to `array-api-strict` to make testing easier. The default device continues to be the CPU device and for arrays that use it nothing should change. However, you can now place an array on a different device with `array_api_strict.Device("pony")` (or some other string, each string is a new device). For arrays on a device that isn't the CPU device calls like `np.asarray(some_strict_array)` will raise an error. This mirrors how PyTorch treats arrays on the CPU and MPS device.

What isn't yet implemented in this PR is raising an error if you try to operate on arrays that are not on the same device.

I wanted to open this PR already now after just a short amount of effort to get feedback what people think about this before putting in the time to update all the tests, etc.

### DIFF
--- a/array_api_strict/__init__.py
+++ b/array_api_strict/__init__.py
@@ -309,6 +309,9 @@ from ._utility_functions import all, any
 
 __all__ += ["all", "any"]
 
+from ._array_object import Device
+__all__ += ["Device"]
+
 # Helper functions that are not part of the standard
 
 from ._flags import (
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -86,7 +86,7 @@ class Array:
     # Use a custom constructor instead of __init__, as manually initializing
     # this class is not supported API.
     @classmethod
-    def _new(cls, x, /, device=None):
+    def _new(cls, x, /, device):
         """
         This is a private method for initializing the array API Array
         object.
@@ -218,11 +218,10 @@ class Array:
     def _check_device(self, other):
         """Check that other is on a device compatible with the current array"""
         if isinstance(other, (int, complex, float, bool)):
-            return other
+            return
         elif isinstance(other, Array):
             if self.device != other.device:
                 raise RuntimeError(f"Arrays from two different devices ({self.device} and {other.device}) can not be combined.")
-            return other
 
     # Helper function to match the type promotion rules in the spec
     def _promote_scalar(self, scalar):
@@ -275,7 +274,7 @@ class Array:
         # behavior for integers within the bounds of the integer dtype.
         # Outside of those bounds we use the default NumPy behavior (either
         # cast or raise OverflowError).
-        return Array._new(np.array(scalar, dtype=self.dtype._np_dtype))
+        return Array._new(np.array(scalar, dtype=self.dtype._np_dtype), device=CPU_DEVICE)
 
     @staticmethod
     def _normalize_two_args(x1, x2) -> Tuple[Array, Array]:
@@ -307,9 +306,9 @@ class Array:
             # performant. broadcast_to(x1._array, x2.shape) is much slower. We
             # could also manually type promote x2, but that is more complicated
             # and about the same performance as this.
-            x1 = Array._new(x1._array[None])
+            x1 = Array._new(x1._array[None], device=x1.device)
         elif x2.ndim == 0 and x1.ndim != 0:
-            x2 = Array._new(x2._array[None])
+            x2 = Array._new(x2._array[None], device=x2.device)
         return (x1, x2)
 
     # Note: A large fraction of allowed indices are disallowed here (see the
@@ -493,13 +492,13 @@ class Array:
         if self.dtype not in _numeric_dtypes:
             raise TypeError("Only numeric dtypes are allowed in __abs__")
         res = self._array.__abs__()
-        return self.__class__._new(res)
+        return self.__class__._new(res, device=self.device)
 
     def __add__(self: Array, other: Union[int, float, Array], /) -> Array:
         """
         Performs the operation __add__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "numeric", "__add__")
         if other is NotImplemented:
             return other
@@ -511,7 +510,7 @@ class Array:
         """
         Performs the operation __and__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "integer or boolean", "__and__")
         if other is NotImplemented:
             return other
@@ -601,7 +600,7 @@ class Array:
         """
         Performs the operation __eq__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         # Even though "all" dtypes are allowed, we still require them to be
         # promotable with each other.
         other = self._check_allowed_dtypes(other, "all", "__eq__")
@@ -627,7 +626,7 @@ class Array:
         """
         Performs the operation __floordiv__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "real numeric", "__floordiv__")
         if other is NotImplemented:
             return other
@@ -639,7 +638,7 @@ class Array:
         """
         Performs the operation __ge__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "real numeric", "__ge__")
         if other is NotImplemented:
             return other
@@ -675,13 +674,13 @@ class Array:
         """
         Performs the operation __gt__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "real numeric", "__gt__")
         if other is NotImplemented:
             return other
         self, other = self._normalize_two_args(self, other)
         res = self._array.__gt__(other._array)
-        return self.__class__._new(res)
+        return self.__class__._new(res, device=other.device)
 
     def __int__(self: Array, /) -> int:
         """
@@ -730,7 +729,7 @@ class Array:
         """
         Performs the operation __le__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "real numeric", "__le__")
         if other is NotImplemented:
             return other
@@ -742,7 +741,7 @@ class Array:
         """
         Performs the operation __lshift__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "integer", "__lshift__")
         if other is NotImplemented:
             return other
@@ -754,7 +753,7 @@ class Array:
         """
         Performs the operation __lt__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "real numeric", "__lt__")
         if other is NotImplemented:
             return other
@@ -766,7 +765,7 @@ class Array:
         """
         Performs the operation __matmul__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         # matmul is not defined for scalars, but without this, we may get
         # the wrong error message from asarray.
         other = self._check_allowed_dtypes(other, "numeric", "__matmul__")
@@ -779,7 +778,7 @@ class Array:
         """
         Performs the operation __mod__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "real numeric", "__mod__")
         if other is NotImplemented:
             return other
@@ -791,7 +790,7 @@ class Array:
         """
         Performs the operation __mul__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "numeric", "__mul__")
         if other is NotImplemented:
             return other
@@ -803,7 +802,7 @@ class Array:
         """
         Performs the operation __ne__.
         """
-        other = self._check_device(other)
+        self._check_device(other)
         other = self._check_allowed_dtypes(other, "all", "__ne__")
         if other is NotImplemented:
             return other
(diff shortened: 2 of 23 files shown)


betatim:
I think the CPU device should be the default. That way code that exists today should keep working and the only people who notice any changes are those who use the pony device.

asmeurer:
This looks good so far. We need to make sure the semantics specified at https://data-apis.org/array-api/latest/design_topics/device_support.html#semantics are followed, namely, disallowing combining arrays from different devices, and making sure that if a function creates a new array based on an existing array that it uses the same device. 

For tests, ideally this would be tested in array-api-tests, but right now device support is not tested at all there. If you just want to add some basic tests here for now, that is fien. 

Finally, there is the `devices` inspection API. https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.devices.html#array_api.info.devices We need to think about how that will work. One option would just be to create a small but fixed number of devices. Or we could add some flags to make it configurable https://data-apis.org/array-api-strict/api.html#array_api_strict.set_array_api_strict_flags

betatim:
I've rebooted my work. The long pause is because I went on holiday :D

I think we have to limit ourselves to a fixed number of devices, otherwise we can't full fill the requirement that the info extension can provide a list of devices. So now you can use the `CPU_DEVICE` and the (creatively named) `device1` and `device2`.

Slowly making progress towards the creation functions and "array combination" functions respecting the device

betatim:
It looks like it would be quite tricky to add device testing to `array-api-tests`. At least at my level of knowledge (unfamiliar with hypothesis and `array-api-tests`). It looks like you'd add a helper, maybe `all_devices`, similar to `all_dtypes` and then use it in the `@given` decorator. The tricky thing is that how to specify a device depends on the library, in a new version of the standard you could use the inspection API to get all devices. So yeah, for now I might add some basic testing here.

asmeurer:
> Do you use ruff or black or something like that for formatting?

There's no autoformatting on this repo. 

> It looks like it would be quite tricky to add device testing to array-api-tests. At least at my level of knowledge (unfamiliar with hypothesis and array-api-tests). It looks like you'd add a helper, maybe all_devices, similar to all_dtypes and then use it in the @given decorator. The tricky thing is that how to specify a device depends on the library, in a new version of the standard you could use the inspection API to get all devices. So yeah, for now I might add some basic testing here.

I think it would have to use the `devices()` function in the inspection API. That would mean that the tests would only work against the newest version of the standard and it would only work against the compat library, but I think that's fine. You'd also probably want to make it optional. 

It's also possible to do some basic testing using the default device, like that `x.device` and `device=` are consistent. 

The annoying thing for the test suite is making sure every function everywhere is passing `device` through properly so that everything gets created on the same device. It would also probably require some upstream fixes to the hypothesis array-api support.

asmeurer:
I think what we need here are just some big parameterized tests combining basic example arrays with different devices across all the APIs. For instance, there's an existing test that checks type promotion and the "no mixing devices" test could look very similar to that.

array_api_strict/_elementwise_functions.py
```diff
@@ -585,21 +629,23 @@ def log10(x: Array, /) -> Array:
     """
     if x.dtype not in _floating_dtypes:
         raise TypeError("Only floating-point dtypes are allowed in log10")
-    return Array._new(np.log10(x._array))
+    return Array._new(np.log10(x._array), device=x.device)
 
 
-def logaddexp(x1: Array, x2: Array) -> Array:
+def logaddexp(x1: Array, x2: Array, /) -> Array:
```
betatim, line 635: I think this was missing/typo.

array_api_strict/tests/test_elementwise_functions.py
```diff
@@ -19,8 +19,16 @@
 
 import pytest
 
+import array_api_strict
+
+
 def nargs(func):
```
betatim, line 25: I modified this so that it works with decorated/wrapped functions as well. `len(getfullargspec(f).args)` returns zero for functions that are decorated with the array API version decorator. From what I can tell from the Python docs this is kind of on purpose/to preserve existing behaviour. `signature()` does the right thing for wrapped functions, but it needs slightly more explicit work to count the arguments.

I think the intention/rule is that `nargs()` counts the number of positional only arguments, which is basically the "number of arrays you need to pass to a elementwise function". I went with the very explicit way of counting the args partially as a way to make it easier for people from the future to understand what `nargs` is meant to do (even if it contains a bug and doesn't actually do what it is meant to do).

array_api_strict/tests/test_elementwise_functions.py
```diff
@@ -91,12 +99,57 @@ def nargs(func):
     "trunc": "real numeric",
 }
 
+
+def test_nargs():
```
betatim, line 103: A short test to make sure `nargs` works but also that all of the functions that we look at have "the right signature" - I found `logaddexp` was missing that trailing `/` when working on `nargs`. So it seems useful to have a "all functions have a reasonable number of arguments" test.

asmeurer: That's good. The array-api test suite doesn't check for positional-only, though it probably could now that it uses inspect.signature.

betatim:
Does someone know more about the failure? It looks like it is not to do with the actual code but with computing the expected shape and that overflowing because the array is of dtype `int8`. How do we fix that?

asmeurer:
Sorry, that is from a new test that I added in the test suite. I guess I didn't catch all the corner cases. You can ignore it for now.

betatim:
In that case, I think, this is ready?!

I tried to modify all the functions that return an array to take into account the `device` of the input. For some I've added tests that check this, but I think really `array-api-tests` should be doing this (check input and output device are consistent). SO maybe adding tests for all functions is not needed.

array_api_strict/__init__.py
```diff
@@ -309,6 +309,9 @@
 
 __all__ += ["all", "any"]
 
+from ._array_object import Device
+__all__ += ["Device"]
```
asmeurer, line 313: I'd rather not add this to the `__init__.py` since it isn't part of the array API. If it is necessary to have some public APIs to create device objects we should make APIs that are more obviously array-api-strict specific (similar to the flags APIs).

array_api_strict/_array_object.py
```diff
@@ -625,19 +661,21 @@ def __getitem__(
         """
         Performs the operation __getitem__.
         """
+        # XXX Does key have to be on the same device? Is there an exception for CPU_DEVICE?
```
asmeurer, line 664: I guess this is underspecified. I would need to test what PyTorch and others do, but I would suspect that an implicit cross-device array `key` is not something that's intended to be supported, since that still would require an implicit device transfer.

array_api_strict/_array_object.py
```diff
@@ -193,6 +215,15 @@ def _check_allowed_dtypes(self, other: bool | int | float | Array, dtype_categor
...
         return other
 
+    def _check_device(self, other):
+        """Check that other is on a device compatible with the current array"""
+        if isinstance(other, (int, complex, float, bool)):
+            return other
+        elif isinstance(other, Array):
+            if self.device != other.device:
+                raise RuntimeError(f"Arrays from two different devices ({self.device} and {other.device}) can not be combined.")
+            return other
```
asmeurer, line 225: I don't see a need for this function to return something. `_check_allowed_dtypes` returns something because it can return `NotImplemented`, but here the return object is always `other`.

array_api_strict/_dtypes.py
```diff
@@ -121,7 +121,7 @@ def __hash__(self):
     "integer": _integer_dtypes,
     "integer or boolean": _integer_or_boolean_dtypes,
     "boolean": _boolean_dtypes,
-    "real floating-point": _floating_dtypes,
+    "real floating-point": _real_floating_dtypes,
```
asmeurer, line 124: Good catch. Is this covered by one of the new tests?

asmeurer:
It's hard to tell just from the diff if you missed anything. Here are all the places in the code that call `_new` without a device keyword

```
$git grep '_new(' | grep -v 'device'
array_api_strict/_array_object.py:        return Array._new(np.array(scalar, dtype=self.dtype._np_dtype))
array_api_strict/_array_object.py:            x1 = Array._new(x1._array[None])
array_api_strict/_array_object.py:            x2 = Array._new(x2._array[None])
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(res)
array_api_strict/_array_object.py:        return self.__class__._new(self._array.T)
array_api_strict/_creation_functions.py:                return Array._new(new_array)
array_api_strict/_creation_functions.py:    return Array._new(np.from_dlpack(x))
array_api_strict/_data_type_functions.py:        Array._new(array) for array in np.broadcast_arrays(*[a._array for a in arrays])
array_api_strict/_data_type_functions.py:    return Array._new(np.broadcast_to(x._array, shape))
array_api_strict/_linalg.py:        U = Array._new(L).mT
array_api_strict/_linalg.py:    return Array._new(L)
array_api_strict/_linalg.py:    return Array._new(np.cross(x1._array, x2._array, axis=axis))
array_api_strict/_linalg.py:    return Array._new(np.linalg.det(x._array))
array_api_strict/_linalg.py:    return Array._new(np.diagonal(x._array, offset=offset, axis1=-2, axis2=-1))
array_api_strict/_linalg.py:    return Array._new(np.linalg.eigvalsh(x._array))
array_api_strict/_linalg.py:    return Array._new(np.linalg.inv(x._array))
array_api_strict/_linalg.py:    return Array._new(np.linalg.norm(x._array, axis=(-2, -1), keepdims=keepdims, ord=ord))
array_api_strict/_linalg.py:    return Array._new(np.linalg.matrix_power(x._array, n))
array_api_strict/_linalg.py:    return Array._new(np.count_nonzero(S > tol, axis=-1))
array_api_strict/_linalg.py:    return Array._new(np.outer(x1._array, x2._array))
array_api_strict/_linalg.py:    return Array._new(np.linalg.pinv(x._array, rcond=rtol))
array_api_strict/_linalg.py:    return Array._new(_solve(x1._array, x2._array))
array_api_strict/_linalg.py:    return Array._new(np.linalg.svd(x._array, compute_uv=False))
array_api_strict/_linalg.py:    return Array._new(np.asarray(np.trace(x._array, offset=offset, axis1=-2, axis2=-1, dtype=dtype)))
array_api_strict/_linalg.py:    res = Array._new(np.linalg.norm(a, axis=_axis, ord=ord))
array_api_strict/_searching_functions.py:    return tuple(Array._new(i) for i in np.nonzero(x._array))
```

So it would be worth double checking all of those.

asmeurer:
As you mentioned, we should indeed be testing most of this in the test suite. However, I'm not really sure how soon that will happen. There's quite a backlog of things to do in the test suite right now, and my current priority is implementing tests for new functions added in 2023.12 or 2024.12 versions of the standard. So some very simple tests here would not hurt. The tests already have a list of two-argument functions which could be reused.

asmeurer:
Were you wanting to add support for devices that don't support certain dtypes, or is that something that we should add in a later pull request?

betatim:
> Were you wanting to add support for devices that don't support certain dtypes, or is that something that we should add in a later pull request?

I'd do that in a separate PR. If only because this one is already quite long and hard to check by looking at the diff.

array_api_strict/_dtypes.py
betatim: Not explicitly. I found this because a test was failing, looked into this and found it. But also it is too long ago for me to remember what exactly it was :-/

array_api_strict/_array_object.py
betatim: Fine for me. I had it return the input just because it would match the API of the other check. Changed now.

asmeurer:
Nice. I feel much better about this after the latest commit making `device` required. 

It looks like another PR I just merged has created a small conflict here, but other than that, I am +1 to merging this.

betatim:
Thanks for taking this over the finish line!

asmeurer:
I opened https://github.com/data-apis/array-api-strict/issues/70 for ideas for further work here.

betatim:
Yes that is the default device (I've not followed the scipy discussion, so I have no idea if this is a good thing?)

Issue #67, opened 2024-10-15
Should we be implementing __array__?


asmeurer:
Right now the array type implements `__array__`, but just as convenience. This was done back when the package was numpy.array_api.

I'm wondering if we should actually remove this. It isn't part of the array API standard, and could just lead people incorrectly thinking it works. And more notably, it is leading to bugs in array_api_strict itself (like #66) where you can forget to use `._array` when passing an array to a NumPy function and it will still work (but maybe not in the same way, if that function also passes the `dtype` argument for instance).

asmeurer:
I tested removing it and found some more issues where weren't using `._array` like in #66. I pushed fixes for this to #68. 

At the same time, if you remove `__array__`, then NumPy converts it to an object array:

```py
>>> import numpy as np
>>> import array_api_strict as xp
>>> np.asarray(xp.asarray([10]))
array(Array([10], dtype=array_api_strict.int64), dtype=object)
```

Or we can explicitly make `__array__` raise an exception, which would make `np.asarray(<array_api_strict Array>)` fail.

asmeurer:
A small hiccup here is that we implicitly use this logic in `asarray` for conversion from lists of 0-D arrays: https://github.com/data-apis/array-api-strict/blob/019d9356d322356bf5a4ef9026ce06c283d00843/array_api_strict/_creation_functions.py#L97

I believe that should be supported in the standard, but the logic in `asarray` seems to check `__array__` first. The docs don't mention any kind of fallback exception that can be used https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method. Maybe there's some internal function in NumPy for converting lists-of-lists that we can reuse?

asmeurer:
Prototype of this without the fix to asarray at #69.

rgommers:
+1 to include a hint about how to fix. Writing that hint is hard - which is a problem.

There are more relevant comments on gh-69, so let's keep the discussion there I'd say.

asmeurer:
So do you think this change should be reverted? I can also issue a warning in `__array__` instead of removing it outright.

rgommers:
> I can also issue a warning in `__array__` instead of removing it outright.

A warning doesn't help, the SciPy test suite will turn those straight into errors. And there'll be either zero or a million warnings probably.

asmeurer:
https://github.com/data-apis/array-api-strict/pull/91

asmeurer:
[...]

PR #69, opened 2024-10-15
Remove __array__

asmeurer:
This makes it raise an exception, since it isn't supported by the standard (if
we just leave it unimplemented, then np.asarray() returns an object dtype
array, which is not good).

There is one issue here from the test suite, which is that this breaks the
logic in asarray() for converting lists of array_api_strict 0-D arrays. I'm
not yet sure what to do about that.

Fixes https://github.com/data-apis/array-api-strict/issues/67.

This is built on #68, which can and should be merged separately (it isn't controversial).

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -125,28 +125,17 @@ class Array:
             mid = np.array2string(self._array, separator=', ', prefix=prefix, suffix=suffix)
         return prefix + mid + suffix
 
-    # This function is not required by the spec, but we implement it here for
-    # convenience so that np.asarray(array_api_strict.Array) will work.
+    # Disallow __array__, meaning calling `np.func()` on an array_api_strict
+    # array will give an error. If we don't explicitly disallow it, NumPy
+    # defaults to creating an object dtype array, which would lead to
+    # confusing error messages at best and surprising bugs at worst.
+    #
+    # The alternative of course is to just support __array__, which is what we
+    # used to do. But this isn't actually supported by the standard, so it can
+    # lead to code assuming np.asarray(other_array) would always work in the
+    # standard.
     def __array__(self, dtype: None | np.dtype[Any] = None, copy: None | bool = None) -> npt.NDArray[Any]:
-        """
-        Warning: this method is NOT part of the array API spec. Implementers
-        of other libraries need not include it, and users should not assume it
-        will be present in other implementations.
-
-        """
-        # copy keyword is new in 2.0.0; for older versions don't use it
-        # retry without that keyword.
-        if np.__version__[0] < '2':
-            return np.asarray(self._array, dtype=dtype)
-        elif np.__version__.startswith('2.0.0-dev0'):
-            # Handle dev version for which we can't know based on version
-            # number whether or not the copy keyword is supported.
-            try:
-                return np.asarray(self._array, dtype=dtype, copy=copy)
-            except TypeError:
-                return np.asarray(self._array, dtype=dtype)
-        else:
-            return np.asarray(self._array, dtype=dtype, copy=copy)
+        raise ValueError("Conversion from an array_api_strict array to a NumPy ndarray is not supported")
 
     # These are various helper functions to make the array behavior match the
     # spec in places where it either deviates from or is more strict than
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -160,8 +160,8 @@ class Array:
         # We have to allow this to be internally enabled as there's no other
         # easy way to parse a list of Array objects in asarray().
         if _allow_array:
-        if self._device != CPU_DEVICE:
-            raise RuntimeError(f"Can not convert array on the '{self._device}' device to a Numpy array.")
+            if self._device != CPU_DEVICE:
+                raise RuntimeError(f"Can not convert array on the '{self._device}' device to a Numpy array.")
             # copy keyword is new in 2.0.0; for older versions don't use it
             # retry without that keyword.
             if np.__version__[0] < '2':
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -53,6 +53,8 @@ CPU_DEVICE = _cpu_device()
 
 _default = object()
 
+_allow_array = False
+
 class Array:
     """
     n-d array object for the array API namespace.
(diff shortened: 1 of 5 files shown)


betatim:
In scikit-learn's tests we convert array API arrays to Numpy arrays in order to compare their values to some reference array. How would we do that if we remove `__array__` here? Other libraries have `.tonumpy()` and the like, but I think `array-api-strict` doesn't? Should we add something like that as part of this removal?

asmeurer:
Testing is a valid use-case. You could use dlpack if you wanted to stay within the standard. Although I'm still not clear if this is a good idea or not.

betatim:
Good point. We already have custom code to convert from the various array libraries to Numpy. Using
```python
x = xp.asarray([1,2,3], device=...)
np.from_dlpack(x)
```
would maybe simplify that code and remove the need for `__array__`.

thomasjpfan (approved):
> Although I'm still not clear if this is a good idea or not.

I think this is a good idea for `array-api-strict`. Libraries that need to do any array conversion should only depend on `__dlpack__`.

(Although this PR is breaking tests in this repo)

asmeurer:
To clarify a note from the OP: this has stopped working

```py
>>> xp.asarray([[xp.asarray(1), xp.asarray(2)]])
  File "/Users/aaronmeurer/Documents/array-api-strict/array_api_strict/_creation_functions.py", line 97, in asarray
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aaronmeurer/Documents/array-api-strict/array_api_strict/_array_object.py", line 138, in __array__
    raise ValueError("Conversion from an array_api_strict array to a NumPy ndarray is not supported")
ValueError: Conversion from an array_api_strict array to a NumPy ndarray is not supported
```

because it was implicitly relying on `np.asarray()` calling `__array__` on nested array_api_strict arrays.

asmeurer:
Thinking about dlpack, should we be doing anything with the device emulation in `__dlpack__`? https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html#dlpack Maybe we should disable `__dlpack__` when the device isn't the default CPU_DEVICE? @betatim

betatim:
I thought dlpack now allows you to move arrays from one device to another? For example to support the common case of "i would like to convert this array from library foo to a numpy array". Is that right? If yes, it could be cool to support this for the "not CPU devices" here. If only because (in scikit-learn) we convert things back to Numpy arrays during tests.

rgommers:
https://github.com/data-apis/array-api-strict/issues/67#issuecomment-2463750454 points out that this change broke a lot of code, and that there is no guidance on what to replace `xp.asarray` with.

In practice, using `from_dlpack` isn't a 1:1 replacement for SciPy et al. because that won't accept sequences. 

> Libraries that need to do any array conversion should only depend on `__dlpack__`.

I think it's probably a good idea in principle to nudge libraries in this direction. But we need to work out the replacement pattern here first. Unless someone has a ready-made suggestion that can be easily implemented in SciPy, scikit-learn & co, we may want to yank the 2.1.1 and 2.1.2 releases. Actually probably a good idea to do so anyway, because this should not land in a bugfix release (xref gh-89).

lucascolley:
the short-term fix for SciPy shouldn't be so bad, unless I'm missing something? Something like

```python
def _asarray(x, xp_in=None, xp_out=None):
	xp_in = array_namespace(x) if xp_in is None else xp_in
	xp_out = array_namespace(x) if xp_out is None else xp_out
	x = xp_in.asarray(x)
	return xp_out.from_dlpack(x)
```

asmeurer:
Just to be clear, this change and the change to pin NumPy are independent of one another (it appears that xarray had issues with the latter). 

I'm happy to revert (not yank) one or both changes if they are causing issues upstream. The `__array__` thing I was unsure about, but at the same time, it *is* the case that `__array__` is not in the standard, so any issues caused by this could also be caused by another array API library that doesn't implement `__array__`. If it's hard to do things, maybe the standard should be updated accordingly?

Regarding depending on NumPy 2.0, this is primarily a maintenance thing, since a good chunk of the code in array-api-strict is just working around array API non-conformance in NumPy 1.26, and can be removed if we only support 2.0. I was under the impression it wouldn't be a huge issue since array-api-strict is a testing only library, but if it's still too soon to do this, we can wait a bit.

rgommers:
I'd for sure revert this change and release 2.1.3, to unbreak SciPy. Then we have actual time to investigate the fix, hopefully some variant like @lucascolley suggested above.



> it _is_ the case that `__array__` is not in the standard, so any issues caused by this could also be caused by another array API library that doesn't implement `__array__`. If it's hard to do things, maybe the standard should be updated accordingly?

Maybe it does need a tweak in the standard, maybe not - but let's think about it a bit more carefully first before reintroducing the removal of `__array__` for a new feature release.

mdhaber:
In principle, it should be about as simple as proposed above. In practice (gh-21890), I'm finding that library support for `from_dlpack` is not quite ready to rely on.

```python3
import numpy as np  # numpy,  2.1.1, py312h49bc9c5_0, conda-forge
import cupy as cp  # cupy, 13.3.0, py312h584ea29_0, conda-forge
import array_api_strict as strict  # array-api-strict, 2.2, pyhd8ed1ab_0, conda-forge
import torch
import jax.numpy as jnp
from scipy._lib._array_api import array_namespace  # SciPy main

xps = [cp, strict, torch, jnp]
a0 = [1, 2, 3]

for xp in xps:
    a = xp.asarray(a0)
    xp_compat = array_namespace(a)

    try:  # transfer to NumPy
        a_np = np.from_dlpack(a, device='cpu')
        print(f'SUCCESS: {xp.__name__} to {np.__name__}')
    except Exception as e:
        a_np = np.asarray(a0)
        print(f'FAILURE: {xp.__name__} to {np.__name__}: {str(e)}')

    try:  # transfer back to original device
        a_xp = xp_compat.from_dlpack(a_np, device=a.device)
        print(f'SUCCESS: {np.__name__} to {xp.__name__}')
    except Exception as e:
        print(f'FAILURE: {np.__name__} to {xp.__name__}: {str(e)}')

# FAILURE: cupy to scipy._lib.array_api_compat.numpy: __dlpack__() got an unexpected keyword argument 'dl_device'
# FAILURE: scipy._lib.array_api_compat.numpy to cupy: cupy._core.dlpack.from_dlpack() takes no keyword arguments
# SUCCESS: array_api_strict to scipy._lib.array_api_compat.numpy
# SUCCESS: scipy._lib.array_api_compat.numpy to array_api_strict
```

asmeurer:
I'm not sure what's going on with the read-only stuff. It may be prudent to push the discussion about that to either numpy or the array-api repo if it is an issue so people who know more about it can chime in about whether something needs to be changed there.

rgommers:
> I'm finding that library support for `from_dlpack` is not quite ready to rely on. So unless I'm misunderstanding, it looks like we'd have to switch between `from_dlpack` and relying on `xp.asarray` working depending on the library.

@mdhaber and I had a chat about this, but to summarize for other readers: `from_dlpack` without the `device` keyword is fine to use, it's broadly supported. That also serves the same purpose as `__array__` and `asarray` (changing to another library without device transfers).

The `device=` keyword is new in the latest version of the standard and requires DLPack 1.0, which doesn't yet have broad support. It provides the new capability to switch to another library's array type _while doing a device transfer_. This isn't related to this issue directly and should not be needed for library code, although it may certainly be useful for both end user code and for testing purposes in libraries.

PR #73, opened 2024-10-21
FIX Use array's device when promoting scalars

betatim:
When promoting a scalar in the context of an existing array, e.g. `x ** 2` we should use the device of `x` for the array we create.

### DIFF
--- a/array_api_strict/tests/test_array_object.py
+++ b/array_api_strict/tests/test_array_object.py
@@ -6,7 +6,7 @@ import numpy as np
 import pytest
 
 from .. import ones, asarray, result_type, all, equal
-from .._array_object import Array, CPU_DEVICE
+from .._array_object import Array, CPU_DEVICE, Device
 from .._dtypes import (
     _all_dtypes,
     _boolean_dtypes,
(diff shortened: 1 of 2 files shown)


array_api_strict/tests/test_array_object.py
```diff
@@ -88,6 +88,14 @@ def test_validate_index():
     assert_raises(IndexError, lambda: a[0])
     assert_raises(IndexError, lambda: a[:])
 
+def test_promoted_scalar_inherits_device():
+    device1 = Device("device1")
+    x = asarray([1., 2, 3], device=device1)
+
+    y = x ** 2
```
betatim, line 95: Without the change this line would fail, but I felt like having the `assert` below is somehow useful for readers from the future, even though it isn't strictly necessary

asmeurer:
@betatim should I make a release with these changes or are there more incoming?

betatim:
I don't know, maybe lets wait a bit. I've been finding these via running the scikit-learn tests, so far they don't run to completion. But it needs a bit of digging to figure out what the bug is each time

PR #74, opened 2024-10-21
Fix the definition of sign() for complex numbers

asmeurer:
This is correct in NumPy 2.0 but not in 1.x.


asmeurer:
The test suite won't actually test this until https://github.com/data-apis/array-api-tests/pull/293 is merged, but I've confirmed the tests pass locally.

PR #78, opened 2024-10-31
Add support for scalar arguments to xp.where

betatim:
This is related to https://github.com/data-apis/array-api/issues/807

This only adds support for scalars to `where()`. If the second or third argument is a scalar it is turned into an array on the same device as `condition`.

I added `2024.12` as `next_supported_version`. This means we can use it in the tests. Not sure if we use some kind of dummy value instead of 2024.12?

### DIFF
--- a/array_api_strict/_searching_functions.py
+++ b/array_api_strict/_searching_functions.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from ._array_object import Array
 from ._dtypes import _result_type, _real_numeric_dtypes
-from ._flags import requires_data_dependent_shapes, requires_api_version
+from ._flags import requires_data_dependent_shapes, requires_api_version, get_array_api_strict_flags
 
 from typing import TYPE_CHECKING
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


asmeurer:
Some of this overlaps with the stuff I've done over at https://github.com/data-apis/array-api-strict/pull/82, although I haven't implemented any scalar stuff there yet because it isn't in the standard yet.

array_api_strict/_searching_functions.py
```diff
@@ -72,17 +72,24 @@ def searchsorted(
...
 
-def where(condition: Array, x1: Array, x2: Array, /) -> Array:
+def where(condition: Array, x1: bool | int | float | Array, x2: bool | int | float | Array, /) -> Array:
     """
     Array API compatible wrapper for :py:func:`np.where <numpy.where>`.
 
     See its docstring for more information.
     """
+    if get_array_api_strict_flags()['api_version'] > '2023.12':
+        if isinstance(x1, (bool, float, int)):
```
ev-br, line 82:
```suggestion
        if isinstance(x1, (bool, float, complex, int)):
```

ev-br (commented):
I'd think it should allow complex `x1` and `x2`, too. Here and in the annotation.

LGTM otherwise!

ev-br:
A small update to make it adhere to the spec to the letter: https://github.com/betatim/array-api-strict/pull/1

betatim:
Thanks for the PR @ev-br  If you'd like to take over the original PR (this PR) to get it done that would be awesome.

ev-br:
Happy to oblige @betatim --- thank you!

PR #82, opened 2024-11-05
Add preliminary support for the draft 2024.12 version of the standard

asmeurer:


asmeurer:
Going to merge and release this. This is all hidden behind a flag and warning, so even though this is still fairly untested, I think it's OK let people start playing with it. 

I believe I have included every change to the draft standard since 2023.12. However, there are still some changes that are planned for 2024.12, but which aren't merged in the standard yet and which I haven't implemented here yet. The main one of interest is scalar support for array functions like `where` (see also #78). My plan is to implement this once there is an implementation in the standard.

PR #83, opened 2024-11-06
2.1.1 release

asmeurer:


asmeurer:
The release script is broken. I'll see if I get a response here https://github.com/pypa/gh-action-pypi-publish/issues/283#issuecomment-2461043762. Otherwise I'm going to just disable attestations (and temporarily disable testpypi) to get this working.

PR #86, opened 2024-11-07
Allow __dlpack__ to work with newer versions of NumPy

### LINKED ISSUE #85: The max_version argument to __dlpack__ is not yet implemented
I see that `__array__` is removed in the latest version, and `from_dlpack` is suggested. But when I use it, I found that `from_dlpack` throws the following error:

```py
>>> import numpy as np
>>> import array_api_strict
>>> np.from_dlpack(array_api_strict.asarray(0))
(1, 0)
  File "/home/jz748/anaconda3/lib/python3.10/site-packages/array_api_strict/_array_object.py", line 592, in __dlpack__
    raise NotImplementedError("The max_version argument to __dlpack__ is not yet implemented")
NotImplementedError: The max_version argument to __dlpack__ is not yet implemented
```

version: numpy 2.1.3 array_api_strict 2.1.1

So, there is no way to convert an array_api_strict array to numpy?

asmeurer:
Fixes #85

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -38,7 +38,7 @@ import types
 
 if TYPE_CHECKING:
     from typing import Optional, Tuple, Union, Any
-    from ._typing import PyCapsule, Device, Dtype
+    from ._typing import PyCapsule, Dtype
     import numpy.typing as npt
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


asmeurer:
I think I'm going to just drop NumPy 1.x support in a subsequent PR.

PR #87, opened 2024-11-08
Require NumPy >= 2.1

### LINKED ISSUE #21: Only support NumPy 2.0 as a backend
Obviously not just yet, but once NumPy 2.0 is released and stable, we should drop support for NumPy 1.26 as a backend. That would allow us to remove all the compatibility code and only have code related to strictness. 

The actual backend in array-api-strict doesn't really matter that much, but this would limit the usability for libraries that use NumPy internally in addition to the array API, so we should only do this once the ecosystem seems generally caught up on NumPy 2.0 support.

asmeurer:
Fixes #21


asmeurer:
Just to be clear for future readers, we can go way beyond what I did here if array-api-strict only supported NumPy 2.0. Virtually all code that is based on compatibility could be removed, with only code based on strictness left.

PR #90, opened 2024-11-08
Revert "Require NumPy >= 2.1"

asmeurer:
Apparently this is causing issues with some libraries (for example, https://github.com/pydata/xarray/issues/9750, because numba doesn't support NumPy 2.0 yet IIUC)

Reverts data-apis/array-api-strict#87


rgommers (approved):
LGTM, thanks @asmeurer! Straight revert seems like a good choice - this can be delayed.

PR #91, opened 2024-11-08
Re-introduce __array__ with a warning

asmeurer:
https://github.com/data-apis/array-api-strict/issues/67

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import operator
 from enum import IntEnum
+import warnings
 
 from ._creation_functions import asarray
 from ._dtypes import (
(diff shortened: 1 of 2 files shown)


asmeurer:
@rgommers mentioned this might still be too hard for SciPy in #67 so waiting for feedback here CC @mdhaber @lucascolley

mdhaber:
I don't know whether silencing the warning is easier than fixing it. @lucascolley do you think we will be able to get scipy/scipy#21835 working before we need the features from `array_api_strict` 2.1+?

lucascolley:
I don't think I have the capacity to push it through immediately, but maybe someone can follow through on the idea.

asmeurer:
OK, I'll just re-enable it completely for now https://github.com/data-apis/array-api-strict/pull/92. We can continue the discussion about it at https://github.com/data-apis/array-api-strict/issues/67. If there are any changes that need to be made for either the standard or NumPy we should discuss those too.

lucascolley:
FWIW I do think it should be removed. But giving consumer libraries some time to adapt will be helpful.

mdhaber:
I'm happy with this being re-enabled while we figure out how this should be done. Maybe it could be controlled by an environment variable so we can test with `_allow_array = False` once we have converted to `from_dlpack`?

asmeurer:
What would help is some clarity on what the problems are. Is there something that should be added to the standard and/or some changes that should be made to NumPy to make this easier? 

I do know that for array-api-strict itself I found that I had to do some workarounds to make `asarray([<list of array_api_strict arrays>])` work, because we reuse the `np.asarray` logic for that (I'm not keen on re-implementing the full list-of-lists logic here, and NumPy doesn't provide any useful helpers). But I don't know if this is the sort of problem that is affecting you, or if the problems are something else.

lucascolley:
In SciPy the problem is that we punted on using DLPack by just using `np.asarray` where needed and adding skips for GPU backends. Now that there is a CPU backend which doesn't work with `np.asarray` (array-api-strict), were seeing the consequences. We do need to move to DLPack, but it will take a bit of time.

mdhaber:
> Is there something that should be added to the standard and/or some changes that should be made to NumPy to make this easier?

We'll probably figure this out when we're making the change to `from_dlpack`. It just wasn't done before because we were supposed to use `np.asarray` to convert to NumPy according to our [development documentation](https://docs.scipy.org/doc/scipy/dev/api-dev/array_api.html).

asmeurer:
Going to discuss `__array__` a little more in the meeting Thursday. For now, it sounds like a warning is no more helpful than an error, so I'm going to close this.

PR #92, opened 2024-11-08
Re-enable __array__

asmeurer:
Removing it caused issues for SciPy
(https://github.com/data-apis/array-api-strict/issues/67). I have left the flag in to make it easy to remove it in the future.

I also considered raising a warning in __array__, but this is also difficult to handle https://github.com/data-apis/array-api-strict/pull/91


asmeurer:
@rgommers @mdhaber @lucascolley

rgommers:
thanks!

Issue #96, opened 2024-11-15
Support for Python scalars


asmeurer:
The 2024.12 version of the standard will implement support for scalars in functions. See https://github.com/data-apis/array-api/issues/807.

This has not yet been written in the draft spec, so nothing is implemented here yet. This issue is just to track that work. 

There has also been a start to this at #78 

See also https://github.com/data-apis/array-api-strict/issues/77

ev-br:
Here are the individual PRs which implement parts of this issue:

- [x] binary elementwise functions: gh-100
- [x] `where`:  gh-78
- [ ] `result_type`: gh-119

PR #100, opened 2024-11-22
ENH: allow python scalars in binary elementwise functions

ev-br:
towards #96

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -230,6 +230,8 @@ class Array:
         elif isinstance(other, Array):
             if self.device != other.device:
                 raise ValueError(f"Arrays from two different devices ({self.device} and {other.device}) can not be combined.")
+        else:
+            raise TypeError(f"Cannot combine an Array with {type(other)}.")
 
     # Helper function to match the type promotion rules in the spec
     def _promote_scalar(self, scalar):
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -230,8 +230,6 @@ class Array:
         elif isinstance(other, Array):
             if self.device != other.device:
                 raise ValueError(f"Arrays from two different devices ({self.device} and {other.device}) can not be combined.")
-        else:
-            raise TypeError(f"Cannot combine an Array with {type(other)}.")
 
     # Helper function to match the type promotion rules in the spec
     def _promote_scalar(self, scalar):
--- a/array_api_strict/_helpers.py
+++ b/array_api_strict/_helpers.py
@@ -1,0 +1,37 @@
+"""Private helper routines.
+"""
+
+from ._flags import get_array_api_strict_flags
+from ._dtypes import _dtype_categories
+
+_py_scalars = (bool, int, float, complex)
+
+
+def _maybe_normalize_py_scalars(x1, x2, dtype_category, func_name):
+
+    flags = get_array_api_strict_flags()
+    if flags["api_version"] < "2024.12":
+        # scalars will fail at the call site
+        return x1, x2
+
+    _allowed_dtypes = _dtype_categories[dtype_category]
+
+    if isinstance(x1, _py_scalars):
+        if isinstance(x2, _py_scalars):
+            raise TypeError(f"Two scalars not allowed, got {type(x1) = } and {type(x2) =}")
+        # x2 must be an array
+        if x2.dtype not in _allowed_dtypes:
+            raise TypeError(f"Only {dtype_category} dtypes are allowed {func_name}. Got {x2.dtype}.")
+        x1 = x2._promote_scalar(x1)
+
+    elif isinstance(x2, _py_scalars):
+        # x1 must be an array
+        if x1.dtype not in _allowed_dtypes:
+            raise TypeError(f"Only {dtype_category} dtypes are allowed {func_name}. Got {x1.dtype}.")
+        x2 = x1._promote_scalar(x2)
+    else:
+        if x1.dtype not in _allowed_dtypes or x2.dtype not in _allowed_dtypes:
+            raise TypeError(f"Only {dtype_category} dtypes are allowed {func_name}. "
+                            f"Got {x1.dtype} and {x2.dtype}.")
+    return x1, x2
+
--- a/array_api_strict/tests/test_array_object.py
+++ b/array_api_strict/tests/test_array_object.py
@@ -96,12 +96,60 @@ def test_promoted_scalar_inherits_device():
 
     assert y.device == device1
 
+
+BIG_INT = int(1e30)
+
+def _check_op_array_scalar(dtypes, a, s, func, func_name, BIG_INT=BIG_INT):
+    # Test array op scalar. From the spec, the following combinations
+    # are supported:
+
+    # - Python bool for a bool array dtype,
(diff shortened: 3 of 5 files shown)


ev-br:
TODO:

- [ ] other functions from https://github.com/data-apis/array-api/issues/807
- [ ] type annotations: currently it's still `add(x1: Array, x2: Array)`; do we want to update (and if yes, how? `ArrayOrScalar` or verbose explicit `Array | bool | int | float | complex`)

ev-br:
cross-ref https://github.com/data-apis/array-api/pull/862

array_api_strict/_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+_py_scalars = (bool, int, float, complex)
+
+
+def _maybe_normalize_py_scalars(x1, x2):
+    from ._array_object import Array
+
+    flags = get_array_api_strict_flags()
+    if flags["api_version"] < "2024.12": # XXX: string comparison for versions
```
asmeurer, line 15: I've been using string comparisons everywhere for versions. It should be fine since every array API version is `202X.12`.

asmeurer:
We need to make sure this properly disallows cross-kind combinations that aren't supported. That will presumably match the text here https://data-apis.org/array-api/latest/API_specification/type_promotion.html#mixing-arrays-with-python-scalars (if https://github.com/data-apis/array-api/issues/841 is implemented, real -> complex will also be allowed).

ev-br:
The whole binop/binary elementwise functions is probably worth a larger rework. ATM there is duplication of logic in matching pairs, like `xp.add` and `Array.__add__`, and the duplicate logic is subtly different, and scalars specifically, array methods seem to be ahead. 

Maybe the right thing to do is to transfer the logic from array methods to elementwise functions and delegate from functions to array methods. There's a precedent in torch dynamo, which just registers ufuncs as ndarray methods:  https://github.com/pytorch/pytorch/blob/main/torch/_numpy/_ndarray.py#L197. 

This nicely avoids duplication and slight incompatibilities.

ev-br:
The CI failure is unrelated: https://github.com/data-apis/array-api-tests/issues/280, hopefully worked around by https://github.com/data-apis/array-api-tests/pull/325

array_api_strict/_array_object.py
```diff
@@ -234,6 +234,8 @@ def _check_device(self, other):
         elif isinstance(other, Array):
             if self.device != other.device:
                 raise ValueError(f"Arrays from two different devices ({self.device} and {other.device}) can not be combined.")
+        else:
+            raise TypeError(f"Cannot combine an Array with {type(other)}.")
```
betatim, line 238: When does it happen that we get here with `other` being neither a scalar nor an array? My expectation was somehow that  this wouldn't happen

Maybe worth adding a comment given that it isn't immediately obvious to me now and probably also not obvious to readers from the future.

array_api_strict/tests/test_array_object.py
```diff
@@ -112,12 +160,12 @@ def test_operators():
...
         "__ne__": "all",
-        "__or__": "integer_or_boolean",
+        "__or__": "integer or boolean",
         "__pow__": "numeric",
         "__rshift__": "integer",
         "__sub__": "numeric",
-        "__truediv__": "floating",
-        "__xor__": "integer_or_boolean",
+        "__truediv__": "floating-point",
+        "__xor__": "integer or boolean",
```
betatim, line 168: For my education, is there a reason for renaming these other than aesthetics? In particular the `integer_or_boolean` makes me wonder if there was a reason why it was created without spaces in it

betatim:
I'm a little undecided on the changes to `_elementwise_functions.py`. It reduces the amount of content in the file and the amount of "dumb" search&replace that needs doing when we want to edit it. This is nice. However the current `main` is super explicit and straight-forward, which is nice if you need to special case something and for understanding what is what. So yeah.

array_api_strict/_array_object.py
ev-br: Good catch, thank you! This is interference from https://github.com/data-apis/array-api-strict/pull/103. 
Should probably decide on that PR one way or another and rebase this one.

array_api_strict/tests/test_array_object.py
ev-br: Totally not aesthetics! This is to match the keys from https://github.com/data-apis/array-api-strict/blob/main/array_api_strict/_dtypes.py#L117 , which are `"integer or boolean"` without underscores. Now that `_check_op_array_scalar` from `test_array_object.py` is reused in `test_elementwise_functions.py` and there is looping over the keys of the `_dtype_categories` dict, it saves a bit of gymnastics to map different spellings onto each other.

ev-br:
Re copy-paste vs generate functions: while I'm definitely not too attached to the current way (it's borderline too meta to my tastes TBH), what I'd be worried about is consistency. Too easy for different functions to drift apart slightly, and we're trying to be super-consistent here, so I'd guess this way will be a little less maintenance going forward. Even at an expense of an initial "what on earth is this" kind of first reaction.

array_api_strict/tests/test_array_object.py
betatim: Aha! That sounds totally sensible, less 🤸

betatim:
Let's see what others think. My main worry with the code generation approach is that we forget to do something which leads to some subtle breakage somewhere in a way that will be tricky to debug. A bit similar to how you should use `functools.wraps` for decorators even though at (very) first glance it seems like it is trivial to do yourself, except it isn't :D

Let's see what opinions and feelings others have.

lucascolley:
> Re copy-paste vs generate functions: while I'm definitely not too attached to the current way (it's borderline too meta to my tastes TBH), what I'd be worried about is consistency. Too easy for different functions to drift apart slightly, and we're trying to be super-consistent here, so I'd guess this way will be a little less maintenance going forward. Even at an expense of an initial "what on earth is this" kind of first reaction.

> Let's see what others think. My main worry with the code generation approach is that we forget to do something which leads to some subtle breakage somewhere in a way that will be tricky to debug. A bit similar to how you should use functools.wraps for decorators even though at (very) first glance it seems like it is trivial to do yourself, except it isn't :D
> 
> Let's see what opinions and feelings others have.

My inclination is towards leaving as is with the straight-forward approach. I think the "what on earth is this" might be a net-negative on the maintenance burden, but of course that depends on how many different people will have to maintain the file. Even if it did increase maintainability slightly I think I agree with Tim that it could lead to bugs which are tricky to debug.

But I don't feel strongly, happy for it to go the other way if people feel so.

rgommers:
I'd go for merging this change. One may be able to argue that there's slightly easier to understand code patterns and the refactor could have been done a bit differently, but it's hard to argue against deduplicating many of the same 5 lines. With this, adding a validation change like `_maybe_normalize_py_scalars` can be done in a single place, and adding a function that is added to the standard is a 1-2 line change. Seems like a win overall; if a bug shows up it shouldn't be very disruptive and probably easily fixable.

betatim:
I'm fine with the decision to use this PR and merge it.

ev-br:
https://github.com/data-apis/array-api/pull/862 has landed, this PR has been okay'd, so lets land this one, too. Thanks to all the reviewers!

PR #101, opened 2024-11-26
ENH: prohibit astype(complex, not complex)

ev-br:
as discussed in https://github.com/data-apis/array-api-tests/pull/311 :

the spec stipulates that (https://data-apis.org/array-api/latest/API_specification/generated/array_api.astype.html#astype)

> Casting a complex floating-point array to a real-valued data type should not be permitted.

Thus error out here and avoid testing `astype(complex, not complex)`  it in `-tests`.


asmeurer (approved):

ev-br:
Merging as approved.

Issue #102, opened 2024-11-27
Arithmetic operations accept numpy arrays


ev-br:
Supposedly, mixing `array-api-strict` arrays with other array types should not be allowed.

Or all of them should be allowed, but then we'd need to specify something like `__array_priority__`  and that opens quite a Pandora box, so I guess not?

```
In [5]: import numpy as np

In [6]: import array_api_strict as xp

In [7]: xp.arange(5, dtype=xp.int8) + np.arange(5, dtype=np.complex64)
Out[7]: array([0.+0.j, 2.+0.j, 4.+0.j, 6.+0.j, 8.+0.j], dtype=complex64)           # xp + np -> np !

In [8]: import torch

In [10]: xp.arange(5, dtype=xp.int8) + torch.arange(5)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 xp.arange(5, dtype=xp.int8) + torch.arange(5)

TypeError: unsupported operand type(s) for +: 'Array' and 'Tensor'

In [11]: import jax.numpy as jnp

In [12]: xp.arange(5, dtype=xp.int8) + jnp.arange(5, dtype=jnp.complex64)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[12], line 1
----> 1 xp.arange(5, dtype=xp.int8) + jnp.arange(5, dtype=jnp.complex64)

TypeError: unsupported operand type(s) for +: 'Array' and 'jaxlib.xla_extension.ArrayImpl'
```

ev-br:
The offender is https://github.com/data-apis/array-api-strict/blob/main/array_api_strict/_array_object.py#L189 :

```
    def _check_allowed_dtypes(self, other: bool | int | float | Array, dtype_category: str, op: str) -> Array:
        ...
        if isinstance(other, (int, complex, float, bool)):
            ...
        elif isinstance(other, Array):
            ....
        else:
            return NotImplemented
```

and then, after `__add__` returns `NotImplemented` , control flow ends in `__array__` which happily call `np.asarray(self._array)`.

ISTM it's best to add an explicit type check to `__op__(self, other)` and explicitly whitelist arrays or scalars, so that we do not depend on https://github.com/data-apis/array-api-strict/issues/67.

asmeurer:
[...]

PR #103, opened 2024-11-27
BUG: add missing `_check_type_device` calls

ev-br:
A minimal fix for gh-102.
As a byproduct, add several `check_device` calls, where previously missing.

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -1104,6 +1104,7 @@ class Array:
         """
         Performs the operation __imod__.
         """
+        self._check_type_device(other)
         other = self._check_allowed_dtypes(other, "real numeric", "__imod__")
         if other is NotImplemented:
             return other
@@ -1126,6 +1127,7 @@ class Array:
         """
         Performs the operation __imul__.
         """
+        self._check_type_device(other)
         other = self._check_allowed_dtypes(other, "numeric", "__imul__")
         if other is NotImplemented:
             return other
@@ -1148,6 +1150,7 @@ class Array:
         """
         Performs the operation __ior__.
         """
+        self._check_type_device(other)
         other = self._check_allowed_dtypes(other, "integer or boolean", "__ior__")
         if other is NotImplemented:
             return other
@@ -1170,6 +1173,7 @@ class Array:
         """
         Performs the operation __ipow__.
         """
+        self._check_type_device(other)
         other = self._check_allowed_dtypes(other, "numeric", "__ipow__")
         if other is NotImplemented:
             return other
@@ -1182,6 +1186,7 @@ class Array:
         """
         from ._elementwise_functions import pow  # type: ignore[attr-defined]
 
+        self._check_type_device(other)
         other = self._check_allowed_dtypes(other, "numeric", "__rpow__")
         if other is NotImplemented:
             return other
@@ -1193,6 +1198,7 @@ class Array:
         """
         Performs the operation __irshift__.
         """
+        self._check_type_device(other)
         other = self._check_allowed_dtypes(other, "integer", "__irshift__")
         if other is NotImplemented:
             return other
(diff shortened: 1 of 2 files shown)


asmeurer:
Should add some tests for this.

ev-br:
I believe https://github.com/data-apis/array-api-strict/pull/103/commits/6d67d4694fa7aea2040729a5071199f5be49321b#diff-8047174d1f9158b619da8e809ed90fd46404caae67e7f687c5ba3a77f5bddc40R215 is a faithful representation of what we can test.

asmeurer:
Ah sorry, I didn't notice that last bit was actually already in the tests. That should be fine. 

This change seems OK. It fixes some of the more egregious uses of `__array__`. Maybe we should also check scipy with this change too (should we add something like that to the CI?), since the last time things broke so spectacularly.

ev-br:
Indeed, scipy's `$ python dev.py test -b array_api_strict` passes locally with this PR. 

W.r.t. CI testing of downstreams, yeah, this might be useful. An immediate downside though is that it'll make a CI run from  under five minutes to > 1/2 hour. It is definitely worth thinking a bit about a larger CI strategy, here or in -compat. Maybe as a cron job or a manual trigger. I feel it should be a separate PR though.

asmeurer:
I would go with whatever's the most pragmatically useful for testing scipy. It would be a good idea to test it before doing releases at the very least, to avoid the issues we had previously.

array_api_strict/_array_object.py
```diff
@@ -234,6 +234,8 @@ def _check_device(self, other):
         elif isinstance(other, Array):
             if self.device != other.device:
                 raise ValueError(f"Arrays from two different devices ({self.device} and {other.device}) can not be combined.")
+        else:
```
asmeurer, line 237: Maybe we should explicitly only reject ndarray. That way NotImplemented can still work on other types (although I honestly don't know how important that is).

ev-br: I'd think that mixing different array objects is a bad idea in general, and is basically impossible to make consistent. Even when it fails, there are just too many possible reasons or failure modes.
So one either buys in duck-typing all the way (and whatever quacking falls out, it does), or say no to anything which is not the same array namespace. If -strict strives to be strict, the latter makes sense, it seems.
That said, I'm happy with whatever you think is best.

```
In [4]: xp.arange(5) + cp.arange(5)
Cell In[4], line 1
----> 1 xp.arange(5) + cp.arange(5)

File cupy/_core/core.pyx:1265, in cupy._core.core._ndarray_base.__add__()

File cupy/_core/_kernel.pyx:1286, in cupy._core._kernel.ufunc.__call__()

File cupy/_core/_kernel.pyx:159, in cupy._core._kernel._preprocess_args()

File cupy/_core/_kernel.pyx:145, in cupy._core._kernel._preprocess_arg()

TypeError: Unsupported type <class 'array_api_strict._array_object.Array'>

In [5]: cp.arange(3) + xp.arange(3)
Cell In[5], line 1
----> 1 cp.arange(3) + xp.arange(3)

File cupy/_core/core.pyx:1269, in cupy._core.core._ndarray_base.__add__()

TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'add'>, '__call__', array([0, 1, 2]), Array([0, 1, 2], dtype=array_api_strict.int64)): 'ndarray', 'Array'

```

lucascolley: > So one either buys in duck-typing all the way (and whatever quacking falls out, it does), or say no to anything which is not the same array namespace. If -strict strives to be strict, the latter makes sense, it seems.

That sounds good to me.

crusaderky:
Superseded by #145 + #148

ev-br:
The main goal of this PR, to reject `xp.asarray(3) + np.asarray(3)`, was in fact fixed by https://github.com/data-apis/array-api-strict/pull/145/

What this PR does now is, in effect, it adds `_check_type_device` to inplace operations and adds a test.
Checking `__iop__` arguments looks like good change to me.

lucascolley (approved):
thanks Evgeni!

PR #104, opened 2024-11-29
BUG: fix import error against Python's optimized mode

neutrinoceros:
reprod:
```
python -OO -c "import array_api_strict"
```


ev-br (approved):
Without commenting on the usefulness of `python -OO`, as long as it exists, this patch LGTM.
Thank you @neutrinoceros

Issue #99, opened 2024-11-21
Arrays created with `from_dlpack` have a different `device`?


mdhaber:
It seems that when a strict array is created using `from_dlpack`, its `device` property is not set as I would expect. Consequently, strict arrays created with `from_dlpack` cannot operate with strict arrays that were not created with `from_dlpack`.

```python3
import array_api_strict as xp
x = xp.asarray([1, 2, 3])
y = xp.from_dlpack(x)
x * y
# ValueError: Arrays from two different devices (array_api_strict.Device('CPU_DEVICE') and <object object at 0x111f86f90>) can not be combined.

x.device  # array_api_strict.Device('CPU_DEVICE')
y.device  # <object at 0x111f86f90>
```

I actually encountered this problem when `from_dlpack` is used on a NumPy array.

```python3
import numpy as np
import array_api_strict as xp
x = xp.asarray([1, 2, 3])
y = xp.from_dlpack(np.asarray([2, 3, 4]))
x * y  # same problem
```

Crossref scipy/scipy#21890

PR #108, opened 2024-12-10
BUG: `from_dlpack`: fix default device

lucascolley:
closes gh-99


lucascolley:
I can confirm that this unblocks SciPy

array_api_strict/tests/test_creation_functions.py
```diff
@@ -236,3 +236,9 @@ def from_dlpack_2023_12(api_version):
     pytest.raises(exception, lambda: from_dlpack(capsule, copy=False))
     pytest.raises(exception, lambda: from_dlpack(capsule, copy=True))
     pytest.raises(exception, lambda: from_dlpack(capsule, copy=None))
+
+
+def test_from_dlpack_default_device():
+    x = asarray([1, 2, 3])
+    y = from_dlpack(x)
+    assert x.device is y.device
```
betatim, line 244: I think this should be
```suggestion
    assert x.device == y.device
```
because you can create `Device()` instances as you want and `Device` isn't a singleton or something like that. So `xp.Device(device="device2") is xp.Device("device2")` is `False`.

I think in this case the test passes because the instance for the default device is reused, but that seems like a "happy accident"/implementation detail?

Tl;DR: there is AFAIK no downside to `==` and it will always work

betatim (commented):
One nitpick about `is` vs `==`.

We could add an assert to check that the device of `x` is actually the default device?

We could add a non regression test that uses a numpy array as argument for `from_dlpack`?

Otherwise looks good to me

lucascolley:
thanks Tim, addressed

betatim (approved):

betatim:
Thanks for the fix and the test!

PR #117, opened 2025-01-02
isdtype() should raise if parameter is not a dtype

crusaderky:
Raise on a typical mistake made by a user that didn't properly read the API documentation (i.e. me):
```python
>>> xp.isdtype(xp.asarray(1.0), xp.float64))
False
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,12 +128,12 @@ ENV/
 env.bak/
 venv.bak/
 
-# Spyder project settings
+# Project settings
+.idea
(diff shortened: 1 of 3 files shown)


ev-br (approved):
There's room to argue that if `something` is not a dtype, `isdtype(something, kind)` should just return False, because... `something` is not a `kind` of a dtype. That would be pythonic etc, but since this project aims to be strict, the change makes sense. 

LGTM, thanks @crusaderky

PR #121, opened 2025-01-29
Define complex scalar <op> float array

ev-br:
This is a companion to https://github.com/data-apis/array-api/pull/871

The volume of workarounds to make tests pass is not bad bad bad, but is probably worth a longer look.
Maybe `array._promote_scalar` needs to grow a flag to tell whether mixing complex and floats is allowed for a specific caller.


ev-br:
https://github.com/data-apis/array-api/pull/871 is merged

PR #129, opened 2025-03-02
CI: update the workflows

ev-br:
- stop testing python 3.9
- start testing python 3.13
- actually test Array API revision 2024.12
- stop testing 2022.12 revision (only test 2023.12 and 2024.12)

### DIFF
--- a/.github/workflows/array-api-tests.yml
+++ b/.github/workflows/array-api-tests.yml
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
         numpy-version: ['1.26', 'dev']
         exclude:
-          - python-version: '3.8'
-            numpy-version: 'dev'
+          - python-version: '3.13'
+            numpy-version: '1.26'
 
     steps:
     - name: Checkout array-api-strict
--- a/.github/workflows/array-api-tests.yml
+++ b/.github/workflows/array-api-tests.yml
@@ -3,8 +3,8 @@ name: Array API Tests
 on: [push, pull_request]
 
 env:
-  PYTEST_ARGS: "-v -rxXfE --ci --hypothesis-disable-deadline --max-examples 200"
-  API_VERSIONS: "2022.12 2023.12 2024.12"
+  PYTEST_ARGS: "-v -rxXfE --hypothesis-disable-deadline --max-examples 200"
+  API_VERSIONS: "2023.12 2024.12"
 
 jobs:
   array-api-tests:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,11 +5,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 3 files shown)


ev-br:
The test is useful for https://github.com/data-apis/array-api-compat/pull/267.

There are several issues with the test itself:

- on pytorch, `count_nonzero` is not implemented for unsigned integer dtypes. Not much we can do about here.
- on jax, the strategy runs into JAX problems with the lack of int64:

```
In [2]: jnp.asarray([2147483648], dtype=jnp.int64)
<ipython-input-2-df19af8d03da>:1: UserWarning: Explicitly requested dtype <class 'jax.numpy.int64'> requested in asarray is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/jax-ml/jax#current-gotchas for more.
  jnp.asarray([2147483648], dtype=jnp.int64)
OverflowError                             Traceback (most recent call last)
<ipython-input-2-df19af8d03da> in ?()
----> 1 jnp.asarray([2147483648], dtype=jnp.int64)

~/.conda/envs/array-api/lib/python3.11/site-packages/jax/_src/numpy/lax_numpy.py in ?(a, dtype, order, copy, device)
   5816                       "Consider using copy=None or copy=True instead.")
   5817   dtypes.check_user_dtype_supported(dtype, "asarray")
   5818   if dtype is not None:
   5819     dtype = dtypes.canonicalize_dtype(dtype, allow_extended_dtype=True)  # type: ignore[assignment]
-> 5820   return array(a, dtype=dtype, copy=bool(copy), order=order, device=device)

~/.conda/envs/array-api/lib/python3.11/site-packages/jax/_src/numpy/lax_numpy.py in ?(object, dtype, copy, order, ndmin, device)
   5634     # TODO(jakevdp): falling back to numpy here fails to overflow for lists
   5635     # containing large integers; see discussion in
   5636     # https://github.com/jax-ml/jax/pull/6047. More correct would be to call
   5637     # coerce_to_array on each leaf, but this may have performance implications.
-> 5638     out = np.asarray(object, dtype=dtype)
   5639   elif isinstance(object, Array):
   5640     assert object.aval is not None
   5641     out = _array_copy(object) if copy else object

OverflowError: Python integer 2147483648 out of bounds for int32
```

Explicitly limiting the range of elements runs into problems with floats: hypothesis howls about not exactly representable numbers and errors out.

So an ideal fix is to cook up a strategy which both limits the range of ints and does the right thing for floats.

For now, I think, I'm going to merge this test soon to unblock  https://github.com/data-apis/array-api-compat/pull/267 (which probably warrants a quick bugfix release of `array-api-compat`, so is better done sooner than later).

ev-br:
@kgryte could you please either 
- remove the "required" status from the python 3.9 CI runs and add it to the new 3.13 runs, or
- give me permissions to do it?

The general "drop python 3.9, numpy 1.21" strategy was discussed in https://github.com/data-apis/array-api-compat/issues/230#issuecomment-2646523051 and I think it's now safe to proceed, both here and in array-api-compat.

lucascolley (approved):

ev-br:
Thanks @kgryte! 

> Ping me again once this PR is merged so I can add the 3.13 rules

ping :-)

kgryte:
@ev-br Updated.

PR #135, opened 2025-03-21
TYP: type annotations

crusaderky:
- Implement thorough static type annotations
- Align type annotations conventions to `array-api-compat` (https://github.com/data-apis/array-api-compat/pull/257)
- Add `py.typed`
- Manually run mypy and fix _most_ typing issues
- Manually run isort (one-off)
- Fix bug where `iinfo` and `finfo` would return a numpy dtype instead of a DType

# Out of scope
- Fix _all_ mypy errors (some highlight material issues that deserve their own PRs)
- Add mypy to CI
- Make pyright happy
- tests module

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -1023,7 +1023,7 @@ class Array:
         res = self._array.__rand__(other._array)
         return self.__class__._new(res, device=self.device)
 
-    def __ifloordiv__(self, other: Array | complex, /) -> Array:
+    def __ifloordiv__(self, other: Array | float, /) -> Array:
         """
         Performs the operation __ifloordiv__.
         """
@@ -1034,7 +1034,7 @@ class Array:
         self._array.__ifloordiv__(other._array)
         return self
 
-    def __rfloordiv__(self, other: Array | complex, /) -> Array:
+    def __rfloordiv__(self, other: Array | float, /) -> Array:
         """
         Performs the operation __rfloordiv__.
         """
@@ -1105,7 +1105,7 @@ class Array:
         self._array.__imod__(other._array)
         return self
 
-    def __rmod__(self, other: Array | complex, /) -> Array:
+    def __rmod__(self, other: Array | float, /) -> Array:
         """
         Performs the operation __rmod__.
         """
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -16,61 +16,61 @@ of ndarray.
 from __future__ import annotations
 
 import operator
+from collections.abc import Iterator
 from enum import IntEnum
-
-from ._creation_functions import asarray
-from ._dtypes import (
-    _DType,
-    _all_dtypes,
-    _boolean_dtypes,
-    _integer_dtypes,
-    _integer_or_boolean_dtypes,
-    _floating_dtypes,
-    _real_floating_dtypes,
-    _complex_floating_dtypes,
-    _numeric_dtypes,
-    _result_type,
-    _dtype_categories,
-    _real_to_complex_map,
-)
-from ._flags import get_array_api_strict_flags, set_array_api_strict_flags
-
-from typing import TYPE_CHECKING, SupportsIndex
-import types
-
-if TYPE_CHECKING:
-    from typing import Optional, Tuple, Union, Any
-    from ._typing import PyCapsule, Dtype
-    import numpy.typing as npt
+from types import EllipsisType, ModuleType
+from typing import Any, Literal, SupportsIndex
 
 import numpy as np
+import numpy.typing as npt
+
+from ._creation_functions import _default, _Default, asarray
+from ._dtypes import (
+    DType,
+    _all_dtypes,
+    _boolean_dtypes,
+    _complex_floating_dtypes,
+    _dtype_categories,
+    _floating_dtypes,
+    _integer_dtypes,
+    _integer_or_boolean_dtypes,
+    _numeric_dtypes,
+    _real_floating_dtypes,
+    _real_to_complex_map,
+    _result_type,
+)
+from ._flags import get_array_api_strict_flags, set_array_api_strict_flags
+from ._typing import PyCapsule
+
 
 class Device:
-    def __init__(self, device="CPU_DEVICE"):
+    _device: str
+    __slots__ = ("_device", "__weakref__")
+
+    def __init__(self, device: str = "CPU_DEVICE"):
         if device not in ("CPU_DEVICE", "device1", "device2"):
             raise ValueError(f"The device '{device}' is not a valid choice.")
         self._device = device
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"array_api_strict.Device('{self._device}')"
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         if not isinstance(other, Device):
             return False
         return self._device == other._device
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(("Device", self._device))
 
 
 CPU_DEVICE = Device()
 ALL_DEVICES = (CPU_DEVICE, Device("device1"), Device("device2"))
 
-_default = object()
-
 # See https://github.com/data-apis/array-api-strict/issues/67 and the comment
 # on __array__ below.
 _allow_array = True
+
 
 class Array:
     """
@@ -88,11 +88,14 @@ class Array:
 
     """
     _array: npt.NDArray[Any]
+    _dtype: DType
+    _device: Device
+    __slots__ = ("_array", "_dtype", "_device", "__weakref__")
 
     # Use a custom constructor instead of __init__, as manually initializing
     # this class is not supported API.
     @classmethod
-    def _new(cls, x, /, device):
+    def _new(cls, x: np.ndarray | np.generic, /, device: Device | None) -> Array:
         """
         This is a private method for initializing the array API Array
         object.
@@ -107,7 +110,7 @@ class Array:
         if isinstance(x, np.generic):
             # Convert the array scalar to a 0-D array
             x = np.asarray(x)
-        _dtype = _DType(x.dtype)
+        _dtype = DType(x.dtype)
         if _dtype not in _all_dtypes:
             raise TypeError(
                 f"The array_api_strict namespace does not support the dtype '{x.dtype}'"
@@ -120,7 +123,7 @@ class Array:
         return obj
 
     # Prevent Array() from working
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls, *args: object, **kwargs: object) -> Array:
         raise TypeError(
             "The array_api_strict Array object should not be instantiated directly. Use an array creation function, such as asarray(), instead."
         )
@@ -128,7 +131,7 @@ class Array:
     # These functions are not required by the spec, but are implemented for
     # the sake of usability.
 
-    def __repr__(self: Array, /) -> str:
+    def __repr__(self) -> str:
         """
         Performs the operation __repr__.
         """
@@ -159,7 +162,9 @@ class Array:
     # This was implemented historically for compatibility, and removing it has
     # caused issues for some libraries (see
     # https://github.com/data-apis/array-api-strict/issues/67).
-    def __array__(self, dtype: None | np.dtype[Any] = None, copy: None | bool = None) -> npt.NDArray[Any]:
+    def __array__(
+        self, dtype: None | np.dtype[Any] = None, copy: None | bool = None
+    ) -> npt.NDArray[Any]:
         # We have to allow this to be internally enabled as there's no other
         # easy way to parse a list of Array objects in asarray().
         if _allow_array:
@@ -184,7 +189,9 @@ class Array:
     # spec in places where it either deviates from or is more strict than
     # NumPy behavior
 
-    def _check_allowed_dtypes(self, other: bool | int | float | Array, dtype_category: str, op: str) -> Array:
+    def _check_allowed_dtypes(
+        self, other: Array | complex, dtype_category: str, op: str
+    ) -> Array:
         """
         Helper function for operators to only allow specific input dtypes
 
@@ -225,16 +232,18 @@ class Array:
 
         return other
 
-    def _check_device(self, other):
+    def _check_device(self, other: Array | complex) -> None:
         """Check that other is on a device compatible with the current array"""
         if isinstance(other, (int, complex, float, bool)):
             return
         elif isinstance(other, Array):
             if self.device != other.device:
                 raise ValueError(f"Arrays from two different devices ({self.device} and {other.device}) can not be combined.")
+        else:
+            raise TypeError(f"Expected Array | python scalar; got {type(other)}")
 
     # Helper function to match the type promotion rules in the spec
-    def _promote_scalar(self, scalar):
+    def _promote_scalar(self, scalar: complex) -> Array:
         """
         Returns a promoted version of a Python scalar appropriate for use with
         operations on self.
@@ -291,7 +300,7 @@ class Array:
         return Array._new(np.array(scalar, dtype=target_dtype._np_dtype), device=self.device)
 
     @staticmethod
-    def _normalize_two_args(x1, x2) -> Tuple[Array, Array]:
+    def _normalize_two_args(x1: Array, x2: Array) -> tuple[Array, Array]:
         """
         Normalize inputs to two arg functions to fix type promotion rules
 
@@ -327,7 +336,17 @@ class Array:
 
     # Note: A large fraction of allowed indices are disallowed here (see the
     # docstring below)
-    def _validate_index(self, key, op="getitem"):
+    def _validate_index(
+        self,
+        key: (
+            int
+            | slice
+            | EllipsisType
+            | tuple[int | slice | EllipsisType | None, ...]
+            | Array
+        ),
+        op: Literal["getitem", "setitem"],
+    ) -> None:
         """
         Validate an index according to the array API.
 
@@ -509,7 +528,7 @@ class Array:
 
     # Everything below this line is required by the spec.
 
-    def __abs__(self: Array, /) -> Array:
+    def __abs__(self) -> Array:
         """
         Performs the operation __abs__.
         """
@@ -518,7 +537,7 @@ class Array:
         res = self._array.__abs__()
         return self.__class__._new(res, device=self.device)
 
-    def __add__(self: Array, other: Union[int, float, Array], /) -> Array:
+    def __add__(self, other: Array | complex, /) -> Array:
         """
         Performs the operation __add__.
         """
@@ -530,7 +549,7 @@ class Array:
         res = self._array.__add__(other._array)
         return self.__class__._new(res, device=self.device)
 
-    def __and__(self: Array, other: Union[int, bool, Array], /) -> Array:
+    def __and__(self, other: Array | int, /) -> Array:
         """
         Performs the operation __and__.
         """
@@ -542,9 +561,7 @@ class Array:
         res = self._array.__and__(other._array)
         return self.__class__._new(res, device=self.device)
 
-    def __array_namespace__(
-        self: Array, /, *, api_version: Optional[str] = None
-    ) -> types.ModuleType:
+    def __array_namespace__(self, /, *, api_version: str | None = None) -> ModuleType:
         """
         Return the array_api_strict namespace corresponding to api_version.
 
@@ -563,7 +580,7 @@ class Array:
         import array_api_strict
         return array_api_strict
 
-    def __bool__(self: Array, /) -> bool:
+    def __bool__(self) -> bool:
         """
         Performs the operation __bool__.
         """
@@ -573,7 +590,7 @@ class Array:
         res = self._array.__bool__()
         return res
 
-    def __complex__(self: Array, /) -> complex:
+    def __complex__(self) -> complex:
         """
         Performs the operation __complex__.
         """
@@ -584,13 +601,13 @@ class Array:
         return res
 
     def __dlpack__(
-        self: Array,
+        self,
         /,
         *,
-        stream: Optional[Union[int, Any]] = None,
-        max_version: Optional[tuple[int, int]] = _default,
-        dl_device: Optional[tuple[IntEnum, int]] = _default,
-        copy: Optional[bool] = _default,
+        stream: Any = None,
+        max_version: tuple[int, int] | None | _Default = _default,
+        dl_device: tuple[IntEnum, int] | None | _Default = _default,
+        copy: bool | None | _Default = _default,
     ) -> PyCapsule:
         """
         Performs the operation __dlpack__.
@@ -622,14 +639,14 @@ class Array:
                 kwargs['copy'] = copy
             return self._array.__dlpack__(**kwargs)
 
-    def __dlpack_device__(self: Array, /) -> Tuple[IntEnum, int]:
+    def __dlpack_device__(self) -> tuple[IntEnum, int]:
         """
         Performs the operation __dlpack_device__.
         """
         # Note: device support is required for this
         return self._array.__dlpack_device__()
 
-    def __eq__(self: Array, other: Union[int, float, bool, Array], /) -> Array:
+    def __eq__(self, other: Array | complex, /) -> Array:  # type: ignore[override]
         """
         Performs the operation __eq__.
         """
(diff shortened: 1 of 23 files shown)


ev-br:
Can ignore python 3.9, cf https://github.com/data-apis/array-api-strict/pull/129

array_api_strict/_creation_functions.py
```diff
@@ -313,12 +305,14 @@ def linspace(
...
-    return Array._new(np.linspace(start, stop, num, dtype=dtype, endpoint=endpoint), device=device)
+    return Array._new(
+        np.linspace(start, stop, num, dtype=_np_dtype(dtype), endpoint=endpoint),
+        device=device,
+    )
 
 
-def meshgrid(*arrays: Array, indexing: str = "xy") -> List[Array]:
+# Note: indexing was 'str' in <=2024.12
+def meshgrid(*arrays: Array, indexing: str = "xy") -> list[Array]:
```
crusaderky, line 315: https://github.com/data-apis/array-api/pull/915

array_api_strict/_data_type_functions.py
```diff
@@ -140,34 +138,32 @@ def finfo(type: Union[Dtype, Array], /) -> finfo_object:
         float(fi.max),
         float(fi.min),
         float(fi.smallest_normal),
-        fi.dtype,
+        DType(fi.dtype),
```
crusaderky, line 141: Fix runtime bug

crusaderky:
> Can ignore python 3.9, cf #129

Unless you plan to merge #129 now, I can't ignore 3.9 as it would make _all_ CI to fail.

crusaderky:
CC @jorenham  @NeilGirdhar

ev-br (commented):
Thanks for the PR Guido!

I went through up to `_linalg.py` before running out of steam for the week. In general it looks great, my main gripe is about annotations for python scalars:

- several look like global search-and-replace : floordiv, `__gt__` just do not accept `complex`
- I'm -1 for changing `bool | int | float` into just `float`.

array_api_strict/_array_object.py
```diff
@@ -16,62 +16,66 @@
...
-import numpy as np
+try:
+    from types import EllipsisType  # Python >=3.10
+except ImportError:
+    EllipsisType = type(Ellipsis)
 
 class Device:
-    def __init__(self, device="CPU_DEVICE"):
+    _device: str
+    __slots__ = ("_device", "__weakref__")
```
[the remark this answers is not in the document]
crusaderky: It's a performance tweak. I can revert if you feel strongly against it.

array_api_strict/_array_object.py
```diff
@@ -16,62 +16,66 @@
...
         if device not in ("CPU_DEVICE", "device1", "device2"):
             raise ValueError(f"The device '{device}' is not a valid choice.")
         self._device = device
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"array_api_strict.Device('{self._device}')"
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
```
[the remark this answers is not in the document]
crusaderky: this is `Device.__eq__`, not `Array.__eq__`

array_api_strict/_array_object.py
```diff
@@ -88,11 +92,14 @@ class Array:
...
     _array: npt.NDArray[Any]
+    _dtype: DType
+    _device: Device
+    __slots__ = ("_array", "_dtype", "_device", "__weakref__")
 
     # Use a custom constructor instead of __init__, as manually initializing
     # this class is not supported API.
     @classmethod
-    def _new(cls, x, /, device):
+    def _new(cls, x: np.ndarray | np.generic, /, device: Device | None) -> Array:
```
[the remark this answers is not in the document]
crusaderky: The implementation explicitly caters for None or omitted. `asarray` leverages this use case e.g. `xp.asarray(1)`

array_api_strict/_array_object.py
```diff
@@ -120,15 +127,15 @@ def _new(cls, x, /, device):
         return obj
 
     # Prevent Array() from working
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls, *args: object, **kwargs: object) -> Array:
```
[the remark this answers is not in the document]
crusaderky: It buys us being able to run mypy with
```
disallow_incomplete_defs = true
disallow_untyped_defs = true
```

array_api_strict/_array_object.py
```diff
@@ -184,7 +193,9 @@ def __array__(self, dtype: None | np.dtype[Any] = None, copy: None | bool = None
     # spec in places where it either deviates from or is more strict than
     # NumPy behavior
 
-    def _check_allowed_dtypes(self, other: bool | int | float | Array, dtype_category: str, op: str) -> Array:
+    def _check_allowed_dtypes(
+        self, other: Array | complex, dtype_category: str, op: str
```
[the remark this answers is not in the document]
NeilGirdhar: > As long as this is for a human reader more than for a linter

I think @jorenham is right that this has a lot of drawbacks.  Once you get used to annotations, you know that complex subsumes all of those options.  If you want to make things clear for novices, why not use a docstring?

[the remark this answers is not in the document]
crusaderky: Here's the point though, none of this is for a human reader that isn't pushing a PR to this repo, save for the public functions in `_flags.py`!

https://data-apis.org/array-api-strict/api.html

array_api_strict/_creation_functions.py
```diff
@@ -201,19 +201,15 @@ def eye(
...
-_default = object()
-
 def from_dlpack(
-    x: object,
+    x: SupportsDLPack,
     /,
     *,
-    device: Optional[Device] = _default,
-    copy: Optional[bool] = _default,
+    device: Device | _Default | None = _default,
```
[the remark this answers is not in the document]
crusaderky: this isn't rendered. However, I'll rename it to `Undef` for clarity.

array_api_strict/_dtypes.py
```diff
@@ -1,19 +1,27 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
crusaderky: This PR is still for python 3.9.
In Python >=3.10, you'll still need it whenever there is a type annotation before the definition, or a conditional import inside `if TYPE_CHECKING:`.

array_api_strict/_array_object.py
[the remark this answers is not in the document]
ev-br: Fine by me, as long as the additional strictness is optional.

array_api_strict/_array_object.py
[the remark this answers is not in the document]
ev-br: No strong feelings here. IME slots as a perf or memory optimization are not worth it. But fine to add them here. If a need for a dunder dict appears (numpy considers adding it to funds, FWIW), we'll reconsider.

array_api_strict/_array_object.py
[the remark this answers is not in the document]
ev-br: Sorry, no. If the tooling is not ready, we don't use it until it is. Please roll it back, here and in the -strict PR.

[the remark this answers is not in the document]
NeilGirdhar: What do you mean by the "the tooling is not ready"?  Python typing will probably never the definition of these basic types, right?

array_api_strict/_linalg.py
```diff
@@ -312,14 +317,14 @@ def slogdet(x: Array, /) -> SlogdetResult:
 
 # To workaround this, the below is the code from np.linalg.solve except
 # only calling solve1 in the exactly 1D case.
-def _solve(a, b):
+def _solve(a: np.ndarray, b: np.ndarray) -> np.ndarray:
     try:
-        from numpy.linalg._linalg import (
+        from numpy.linalg._linalg import (  # type: ignore[attr-defined]
```
jorenham, line 322: I opened https://github.com/numpy/numtype/issues/367 for this

array_api_strict/_linalg.py
```diff
@@ -1,33 +1,31 @@
...
 from ._elementwise_functions import conj
-from ._array_object import Array
-from ._flags import requires_extension, get_array_api_strict_flags
+from ._flags import get_array_api_strict_flags, requires_extension
+from ._manipulation_functions import reshape
+from ._statistical_functions import _np_dtype_sumprod
 
 try:
-    from numpy._core.numeric import normalize_axis_tuple
+    from numpy._core.numeric import normalize_axis_tuple  # type: ignore[attr-defined]
```
jorenham, line 25: I opened https://github.com/numpy/numtype/issues/368 for this

array_api_strict/_sorting_functions.py
```diff
@@ -18,7 +20,7 @@ def argsort(
     if x.dtype not in _real_numeric_dtypes:
         raise TypeError("Only real numeric dtypes are allowed in argsort")
     # Note: this keyword argument is different, and the default is different.
-    kind = "stable" if stable else "quicksort"
+    kind: Literal["stable", "quicksort"] = "stable" if stable else "quicksort"
```
jorenham, line 23: FYI, `numpy.sort` only considers the first letter of `kind` case-insensitively, so `"Q"` or `"QuantumSort300"` would also work. 

But since that can't be typed by something narrower than `str`, it currently only allows `{"Q", "quick", "quicksort"}` for quicksort and `{"S", "stable", "stablesort"}` for stable<sub>sort</sub>.

So for the sake of consistency, symmetry, and that vein on my forehead that's about to burst after seeing this, you might also want to consider something like

```suggestion
    kind: Literal["stablesort", "quicksort"] = "stablesort" if stable else "quicksort"
```

*entweder*

```suggestion
    kind: Literal["stable", "quick"] = "stable" if stable else "quick"
```

*soit*

```suggestion
    kind: Literal["S", "Q"] = "S" if stable else "Q"
```

array_api_strict/_sorting_functions.py
```diff
@@ -47,8 +50,7 @@ def sort(
     if x.dtype not in _real_numeric_dtypes:
         raise TypeError("Only real numeric dtypes are allowed in sort")
     # Note: this keyword argument is different, and the default is different.
-    kind = "stable" if stable else "quicksort"
-    res = np.sort(x._array, axis=axis, kind=kind)
+    res = np.sort(x._array, axis=axis, kind="stable" if stable else "quicksort")
```
jorenham, line 53: me repito

array_api_strict/_typing.py
```diff
@@ -8,82 +8,66 @@
...
+        "complex64": DType,
+        "complex128": DType,
+        "int8": DType,
+        "int16": DType,
+        "int32": DType,
+        "int64": DType,
+        "uint8": DType,
+        "uint16": DType,
+        "uint32": DType,
+        "uint64": DType,
```
jorenham, line 70: if there are no spaces, the general recommendation is to use the class-based syntax instead

```py
class DataTypes(TypedDict, total=false):
	bool: DType
	float32: DType
	float64: DType
	complex64: DType
	complex128: DType
	int8: DType
	int16: DType
	int32: DType
	int64: DType
	uint8: DType
	uint16: DType
	uint32: DType
	uint64: DType
```

I think there's something in `ruff` that would complain about it, as well 🤷🏻

array_api_strict/_utility_functions.py
```diff
@@ -57,7 +56,7 @@ def diff(
     # currently specified.
 
     # NumPy does not support prepend=None or append=None
-    kwargs = dict(axis=axis, n=n)
+    kwargs: dict[str, Any] = dict(axis=axis, n=n)
```
jorenham, line 59: The cool kids are all constructing their dicts as `{...}` instead of `dict(...)` nowadays, so how about:

```suggestion
    kwargs: dict[str, int | npt.NDArray[Any]] = {"axis": axis, "n", n}
```

I sneakily narrowed the dict type as well, which should get you couple of extra swag-creds

array_api_strict/tests/test_validation.py
```diff
@@ -1,11 +1,9 @@
-from typing import Callable
-
 import pytest
 
 import array_api_strict as xp
 
 
-def p(func: Callable, *args, **kwargs):
+def p(func, *args, **kwargs):
```
jorenham, line 6: Something like 

```suggestion
def p(func: Callable[..., object], *args: object, **kwargs: object) -> pytest.param:
```

seems just as easy to me 🤷🏻.

But that being said, I'm the type that would even stub his own toe.

pyproject.toml
```diff
@@ -31,3 +31,16 @@ Repository = "https://github.com/data-apis/array-api-strict"
...
 
+[tool.mypy]
+disallow_incomplete_defs = true
+disallow_untyped_decorators = true
+disallow_untyped_defs = true
+no_implicit_optional = true
+show_error_codes = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_unreachable = true
```
jorenham, line 42: I would also recommend `strict_bytes = true`, which disables the "promotion" of `memoryview` and `bytesarray` to `bytes` (similar to how `int` gets promoted to `float` and `float` to `complex`). This will be enabled by default in mypy 2.0.

It might also be a good idea to set `local_partial_types = true`, which will also be the default in mypy 2.0.

Setting `strict = true` is also a good idea (and it isn't as strict as the name suggests it is).

jorenham (commented):
Good stuff! I left a couple of suggestions; do with it what you want 👌🏻

array_api_strict/_helpers.py
```diff
@@ -1,18 +1,24 @@
...
+    x1: Array | complex,
+    x2: Array | complex,
+    dtype_category: str,
+    func_name: str,
+) -> tuple[Array, Array]:
     flags = get_array_api_strict_flags()
     if flags["api_version"] < "2024.12":
         # scalars will fail at the call site
-        return x1, x2
+        return x1, x2  # type: ignore[return-value]
```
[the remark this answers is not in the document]
crusaderky: cast would be wrong. We're deliberately returning a potentially wrong value with the intent to make the caller crash. cast is for when you know the variable is of a certain type but the linter doesn't know yet.

array_api_strict/_info.py
```diff
@@ -36,104 +33,104 @@ def capabilities(self) -> Capabilities:
...
         }
 
     @requires_api_version('2023.12')
     def dtypes(
         self,
         *,
-        device: Optional[device] = None,
-        kind: Optional[Union[str, Tuple[str, ...]]] = None,
+        device: Device | None = None,
+        kind: str | tuple[str, ...] | None = None,
```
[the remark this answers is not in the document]
crusaderky: I'd rather not deviate from the API standard.

array_api_strict/_linalg.py
crusaderky: Aren't they going to answer that you're not meant to import private API?

array_api_strict/_sorting_functions.py
crusaderky: The numpy API specification says 

> kind{‘quicksort’, ‘mergesort’, ‘heapsort’, ‘stable’}, optional

array_api_strict/tests/test_validation.py
crusaderky: Hard no from me. This is a test module.

pyproject.toml
crusaderky: I'd rather treat these settings as a good starting point (I copy-pasted them from my template project) and leave tweaking them to a follow-up PR. There are many unfixed issues with them already.

jorenham: I don't expect that enabling `strict_bytes` and `local_partial_types` will lead to different outcomes in this project.

array_api_strict/tests/test_validation.py
jorenham: > This is a test module.

Yea I noticed that.

> Hard no from me.

Why?

array_api_strict/_info.py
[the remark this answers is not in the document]
jorenham: The API explicitly states that if `kind` is a string, is must be one of the listed string literals. In that sense, using a literal would make it deviate *less* from the API standard.
https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.dtypes.html

array_api_strict/_linalg.py
jorenham: "They" in this context refers to a guy that's very sympathetic to the array API projects <sub>who's also very handsome and has a great sense of humor</sub>.

Coincidentally, "they" just merged a PR that fixes this:

https://github.com/numpy/numtype/pull/371

array_api_strict/_sorting_functions.py
jorenham: Yes, but that's incomplete:

```pycon
>>> import numpy as np
>>> np.sort([1,3,2], kind="S")
array([1, 2, 3])
>>> np.sort([1,3,2], kind="Q")
array([1, 2, 3])
```

array_api_strict/_array_object.py
```diff
@@ -16,62 +16,67 @@
...
-if TYPE_CHECKING:
-    from typing import Optional, Tuple, Union, Any
-    from ._typing import PyCapsule, Dtype
-    import numpy.typing as npt
-
-import numpy as np
 
 class Device:
-    def __init__(self, device="CPU_DEVICE"):
+    _device: str
```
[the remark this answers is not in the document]
crusaderky: I see no reason to prevent another library from subclassing this

array_api_strict/_helpers.py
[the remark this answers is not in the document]
crusaderky: cast is still wrong unless I also update the return type of the function.
Which I'd rather not do, as I'd rather not complicate typing annotations for the benefit of obsolete versions of the API.

array_api_strict/_info.py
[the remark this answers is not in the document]
crusaderky: This must match `isdtype`, which is a _very_ widely used function. This is a nuanced discussion that I'd rather leave to a follow-up PR.

array_api_strict/tests/test_validation.py
crusaderky: type annotations in test modules unnecessarily increase attrition and code verbosity without giving any material benefit to code maintainability or robustness. A pathological example of that is array-api-extra.

ev-br (dismissed):
Let me record my -1 on `bool | int |float` until https://github.com/data-apis/array-api-compat/pull/257#issuecomment-2746365323 is done.

ev-br:
To repeat, my only strong feelings are on 
- consistency of annotations and signatures with the Array API docs and stubs, and
- 'int | bool | float',
in this order.

With these two items being there, I'm declaring my expertise on typing to run out, the typing itself being considered mostly harmless, I'm dismissing my request for changes and bowing out. As long as my boundary conditions are not exceeded, I'm happy with whatever y'all converge on.

array_api_strict/_helpers.py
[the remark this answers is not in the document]
jorenham: fair enough

array_api_strict/_sorting_functions.py
jorenham: I'd agree with you if it would've been consistent on its own, which it isn't in this case.

crusaderky: I can change to `kind: Literal['quicksort', 'mergesort', 'heapsort', 'stable']` if you prefer

array_api_strict/tests/test_validation.py
jorenham: I don't really agree with that. 

Tests have to be maintained as well, and annotations make it easier for the maintainer to see what the *intent* of the tests is. 

In case of test failures, it can be very helpful to see the annotations printed in the error output, because in many cases it is helpful to know the *intent* of the tests when you are figuring out when it fails.

So I guess I view type annotations in tests as a concise yet precise form of documentation, specifically meant to make our own (i.e. the maintainers) lives a bit easier.

Either way, do whatever floats your boat. I just know that it has helped in my projects, so I think it can help here as well.

array_api_strict/_sorting_functions.py
jorenham: if you remove the `*sort` bits, then your got yourself a deal 😉

crusaderky: I find it would be bonkers to prefer an undocumented variant, which potentially may become deprecated in the future, to the official one.

The [official API documentation](https://numpy.org/doc/2.1/reference/generated/numpy.argsort.html) states:

> kind{‘quicksort’, ‘mergesort’, ‘heapsort’, ‘stable’}, optional
> Sorting algorithm. The default is ‘quicksort’. Note that both ‘stable’ and ‘mergesort’ use timsort under the covers and, in general, the actual implementation will vary with data type. The ‘mergesort’ option is retained for backwards compatibility.

There is no mention whatsoever of the other variants accepted in [_SortKind](https://github.com/numpy/numpy/blob/9389862162bbd46b5324402d6a08a27bc18ddb7d/numpy/__init__.pyi#L993-L998), which just happen to be accepted I assume out of backwards compatibility concerns.

jorenham: > I find it would be bonkers to prefer an undocumented variant, which potentially may become deprecated in the future, to the official one.

Well, it's no secret that I'm bonkers, but I'm not sure how that's relevant here 🤷🏻 

And yes, it might become deprecated, just like anything that's documented might.

> The [official API documentation](https://numpy.org/doc/2.1/reference/generated/numpy.argsort.html) states:

The [*other* official API documentation](https://github.com/numpy/numpy/blob/main/numpy/__init__.pyi#L991-L998), often referred to as the stubs, states that it *is* supported: 

```pyi
# in practice, only the first case-insensitive character is considered (so e.g.
# "QuantumSort3000" will be interpreted as quicksort).
_SortKind: TypeAlias = L[
    "Q", "quick", "quicksort",
    "M", "merge", "mergesort",
    "H", "heap", "heapsort",
    "S", "stable", "stablesort",
]
```

But it's also fine by me if you do keep it like this. Because as I've said before, it's  just a suggestion. No need to worry about that vein on my forehead ;)

crusaderky: Stubs are *not* documentation.
Stubs are an optional clutch to detach type annotations from the code. They should otherwise be treated as part of the implementation.

_The_ (one and only) API documentation does not mention those extra possible values. Which means they are an implementation detail.

crusaderky:
I believe there should not be any oustanding comments?

array_api_strict/_sorting_functions.py
jorenham: I disagree (but I'm probably the most biased person you could find). And since there's no "agree to disagree" button, marking it as "resolved" is fine by me

lucascolley:
### Trying to push things forward hat on:

It sounds like both @crusaderky and I would be happy to keep `int | bool | float` in for now given @ev-br's concern, with a view to eventually removing them once we have array-api-typing up and running and readable/documented aliases which make sense appearing both in the 'real' annotations and on the spec pages.

Would merging @crusaderky's 2 PRs disrupt your work at https://github.com/data-apis/array-api-compat/pull/288 @jorenham ?


> - consistency of annotations and signatures with the Array API docs and stubs, and
> - `int | bool | float`,
> in this order.

I haven't read the entirety of all of these threads, but are there any outstanding concerns about consistency *apart from* the `int | bool | float` situation?

---

### Opinion hat on:

For me, there is no significant consistency concern about `int | bool | float`, for reasons already stressed in this thread: these annotations are tool/developer facing, not user facing. I think it is fairly obvious to developers adopting the spec that the spec pages are the source of truth. Following SciPy's lead, we should distinguish between documented types (for docs readability) and the actual type annotations, keeping in mind that we can hope for these to converge somewhat with array-api-typing. In this sense some 'inconsistency' is actually desirable.

If the concern is that inconsistency between the spec saying `int | float` and an annotation in (say) array-api-strict saying `float` will confuse array-api-strict developers, I think this is best remedied in the relevant contributor guides, rather than holding up the typing experts from making the much-desired improvements.

array_api_strict/_array_object.py
```diff
@@ -680,14 +701,14 @@ def __ge__(self: Array, other: Union[int, float, Array], /) -> Array:
...
-            Array,
-        ],
+        self,
+        key: (
+            int
+            | slice
+            | EllipsisType
+            | tuple[int | slice | EllipsisType | None, ...]
+            | Array
+        ),
```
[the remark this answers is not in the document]
crusaderky: Fixed

jorenham:
> Would merging @crusaderky's 2 PRs disrupt your work at [data-apis/array-api-compat#288](https://github.com/data-apis/array-api-compat/pull/288) @jorenham ?

maybe a little bit; but I'll survive

lucascolley:
okay, let's make some progress here then - thanks as always Joren, and Guido for the PR

PR #145, opened 2025-04-02
ENH: disallow numpy generics

crusaderky:
In #135, binary operators vs. numpy.int64 became (correctly) disallowed.
Add a test for it.
Also test binary operators vs. numpy.float64 and numpy.complex128, which must be ~allowed as~ rejected even if they are subclasses of float and complex respectively.

The same treatment for functions is in a separate PR: https://github.com/data-apis/array-api-strict/pull/148.

- XREF https://github.com/scipy/scipy/pull/22783

### DIFF
--- a/array_api_strict/tests/test_array_object.py
+++ b/array_api_strict/tests/test_array_object.py
@@ -255,30 +255,37 @@ def _check_op_array_scalar(dtypes, a, s, func, func_name, BIG_INT=BIG_INT):
             func(s)
         return False
 
+binary_op_dtypes = {
+    "__add__": "numeric",
+    "__and__": "integer or boolean",
+    "__eq__": "all",
+    "__floordiv__": "real numeric",
+    "__ge__": "real numeric",
+    "__gt__": "real numeric",
+    "__le__": "real numeric",
+    "__lshift__": "integer",
+    "__lt__": "real numeric",
+    "__mod__": "real numeric",
+    "__mul__": "numeric",
+    "__ne__": "all",
+    "__or__": "integer or boolean",
+    "__pow__": "numeric",
+    "__rshift__": "integer",
+    "__sub__": "numeric",
+    "__truediv__": "floating-point",
+    "__xor__": "integer or boolean",
+}
+unary_op_dtypes = {
+    "__abs__": "numeric",
+    "__invert__": "integer or boolean",
+    "__neg__": "numeric",
+    "__pos__": "numeric",
+}
 
 def test_operators():
     # For every operator, we test that it works for the required type
     # combinations and raises TypeError otherwise
-    binary_op_dtypes = {
-        "__add__": "numeric",
-        "__and__": "integer or boolean",
-        "__eq__": "all",
-        "__floordiv__": "real numeric",
-        "__ge__": "real numeric",
-        "__gt__": "real numeric",
-        "__le__": "real numeric",
-        "__lshift__": "integer",
-        "__lt__": "real numeric",
-        "__mod__": "real numeric",
-        "__mul__": "numeric",
-        "__ne__": "all",
-        "__or__": "integer or boolean",
-        "__pow__": "numeric",
-        "__rshift__": "integer",
-        "__sub__": "numeric",
-        "__truediv__": "floating-point",
-        "__xor__": "integer or boolean",
-    }
+
     # Recompute each time because of in-place ops
     def _array_vals():
         for d in _integer_dtypes:
@@ -337,12 +344,6 @@ def test_operators():
                             else:
                                 assert_raises(TypeError, lambda: getattr(x, _op)(y))
 
-    unary_op_dtypes = {
-        "__abs__": "numeric",
-        "__invert__": "integer or boolean",
-        "__neg__": "numeric",
-        "__pos__": "numeric",
-    }
     for op, dtypes in unary_op_dtypes.items():
         for a in _array_vals():
             if (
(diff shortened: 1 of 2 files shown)


array_api_strict/tests/test_array_object.py
```diff
@@ -410,6 +411,78 @@ def _matmul_array_vals():
...
+@pytest.mark.parametrize(
+    "op",
+    [
+        op for op, dtypes in binary_op_dtypes.items() 
+        if dtypes not in ("integer", "integer or boolean")
+    ],
+)
+def test_binary_operators_vs_numpy_float(op):
+    """
+    np.float64 is a subclass of float and must be allowed.
```
ev-br, line 438: Have to admit I'm not a fan of this. This is technically correct, it is a subclass. That said, it is an implementation detail IMO. 

I'd think that all instances where numpy returns a numpy scalar, all other libraries return 0D arrays, so I'd think we better always handle them together with arrays not python scalars.

ev-br: Definitely not crash :-).

I think the following is correct[ because `np.float64(2)` is conceptually a `np.asarray(2.)` not python scalar 2.)

```
In [4]: xp = array_namespace(np.empty(2))

In [5]: xp.maximum(xp.asarray(1, dtype=xp.float32), 2)
Out[5]: np.float32(2.0)

In [6]: xp.maximum(xp.asarray(1, dtype=xp.float32), np.float64(2))
Out[6]: np.float64(2.0)
```

Likewise, the following is also correct, for the same reason:

```
In [12]: xpt = array_namespace(torch.empty(2))

In [13]: xpt.maximum(xpt.asarray(1, dtype=xpt.float32), np.float64(2))
Cell In[13], line 1
----> 1 xpt.maximum(xpt.asarray(1, dtype=xpt.float32), np.float64(2))

File ~/repos/array-api-compat/array_api_compat/torch/_aliases.py:91, in _two_arg.<locals>._f(x1, x2, **kwargs)
     88 @_wraps(f)
     89 def _f(x1, x2, /, **kwargs):
     90     x1, x2 = _fix_promotion(x1, x2)
---> 91     return f(x1, x2, **kwargs)

TypeError: maximum(): argument 'other' (position 2) must be Tensor, not numpy.float64
```

crusaderky: > In [6]: xp.maximum(xp.asarray(1, dtype=xp.float32), np.float64(2))
> Out[6]: np.float64(2.0)

This looks definitely wrong to me. array-api-strict shouldn't quietly accept a numpy array.
Right now the output is the expected one without explicitly disallowing inheritance:

```python
>>> xp.maximum(xp.asarray(1, dtype=xp.float32), np.float64(2))
Array(2., dtype=array_api_strict.float32)
>>> xp.maximum(xp.asarray(1, dtype=xp.float32), np.asarray(2, dtype=np.float64))
TypeError: Only real numeric dtypes are allowed in maximum(...). Got array_api_strict.float32 and float64.
>>> xp.maximum(xp.asarray(1, dtype=xp.int32), np.int64(2))
TypeError: Only real numeric dtypes are allowed in maximum(...). Got array_api_strict.int32 and int64.
```

ev-br: > This looks definitely wrong to me. array-api-strict shouldn't quietly accept a numpy array.
> Right now the output is the expected one without explicitly disallowing inheritance:

Yes, in my example `xp` is `array_api_compat.numpy`. As your example shows, array-api-strict does the right thing in main, too. So what is this PR fixing?

array_api_strict/tests/test_array_object.py
```diff
@@ -255,30 +255,37 @@ def _check_op_array_scalar(dtypes, a, s, func, func_name, BIG_INT=BIG_INT):
...
+    "__lt__": "real numeric",
+    "__mod__": "real numeric",
+    "__mul__": "numeric",
+    "__ne__": "all",
+    "__or__": "integer or boolean",
+    "__pow__": "numeric",
+    "__rshift__": "integer",
+    "__sub__": "numeric",
+    "__truediv__": "floating-point",
+    "__xor__": "integer or boolean",
```
ev-br, line 276: Note to self: the values here and in `unary_op_types` below correspond to dtype categories from `_dtypes.py`, so these two dicts are similar to those in https://github.com/data-apis/array-api-strict/blob/main/array_api_strict/tests/test_elementwise_functions.py#L36

array_api_strict/tests/test_array_object.py
```diff
@@ -410,6 +411,78 @@ def _matmul_array_vals():
...
+        op for op, dtypes in binary_op_dtypes.items()
+        if dtypes not in ("real numeric", "floating-point")
+    ],
+)
+def test_binary_operators_vs_numpy_int(op):
+    """np.int64 is not a subclass of int and must be disallowed"""
+    a = asarray(1)
+    i64 = np.int64(1)
+    with pytest.raises(TypeError, match="Expected Array or Python scalar"):
+        getattr(a, op)(i64)
```
ev-br, line 426: Note to self: this test basically tests the status quo, the behavior is the same on main.

array_api_strict/tests/test_array_object.py
```diff
@@ -410,6 +411,78 @@ def _matmul_array_vals():
...
+    np.float32 is not and must be rejected.
+    """
+    a = asarray(1.)
+    f64 = np.float64(1.)
+    f32 = np.float32(1.)
+    func = getattr(a, op)
+    for op in binary_op_dtypes:
+        assert isinstance(func(f64), Array)
+        with pytest.raises(TypeError, match="Expected Array or Python scalar"):
+            func(f32)
```
ev-br, line 448: This test also reflects the status quo: the behavior is exactly the same on main.

array_api_strict/tests/test_array_object.py
```diff
@@ -410,6 +411,78 @@ def _matmul_array_vals():
...
+def test_binary_operators_device_mismatch(op, dtypes):
+    if dtypes in ("real numeric", "floating-point"):
+        dtype = float64
+    else:
+        dtype = int64
+    
+    a = asarray(1, dtype=dtype, device=CPU_DEVICE)
+    b = asarray(1, dtype=dtype, device=Device("device1"))
+    with pytest.raises(ValueError, match="different devices"):
+        getattr(a, op)(b)
```
ev-br, line 483: This test would be great to extend to binary functions, too. I'd expect them to work just the same, because they reuse the `._promote_scalar` logic. If `xp.add(a, b)` differs from `a.__add__(b)`, it's a bug, so would be great to test for.

crusaderky: It turns out extending to binary functions caused the scope blow up, so I moved it to a separate PR: #148

crusaderky:
I've reworked this PR to disallow np.float64 and np.complex128 from being treated as float and complex in Array methods.
The same treatment for functions is in #148.
This is ready for a second round of review.

crusaderky:
> There is a non-zero chance that this causes issues downstream. However I believe that the change is correct and useful and it's downstream that should adjust to better align with the spec.

Well, the whole purpose of array-api-strict is to cause issues early during testing.

ev-br:
> Well, the whole purpose of array-api-strict is to cause issues early during testing.

Exactly. My only point is that if the patch proves too disruptive downstream, we can temporarily revert to give upstream time to adapt. Happened with `__array__` already.

mdhaber:
Haven't worked on `marray` for a while, but I think this broke all reflected operations. I'm not seeing a way around it because, for instance, `__add__` raises before an `marray` __radd__` could be called. Is that supposed to happen?

crusaderky:
> `__add__` raises before an `marray` `__radd__` could be called. Is that supposed to happen?

Yes.

array-api-strict, by design, explicitly disallows everything that in the Array API standard is left as undefined.
This includes interoperability between different Array API-compatible libraries outside of the two APIs that explicitly allow for it, which are `__buffer__` -> `asarray` and `__dlpack__` -> `from_dlpack`.

Adding a naked array-api-strict object to a marray that wraps around array-api-strict falls neatly in the "undefined in Arrayy API, therefore forbidden in array-api-strict" category. If marray supports this operation, great, but it should not expect another arbitrary Array API library to do the same.

The expectation in libraries such as scipy is that all your inputs use the same Array API library.

PR #146, opened 2025-04-03
Fix indexing with integers

### LINKED ISSUE #144: Regression: indexing by `np.int64` indices
The Array API [states](https://data-apis.org/array-api/latest/API_specification/indexing.html):

> An integer index must be an object satisfying [operator.index](https://www.python.org/dev/peps/pep-0357/) (e.g., int).

Therefore, this is legal, because `operator.index(np.int64(0))` works:
```python
>>> import numpy as np
>>> a = xp.asarray([10, 20])
>>> a[np.int64(0)]
10
```

#139 broke it and caused a regression in scipy:
```
FAILED scipy/cluster/tests/test_vq.py::TestKMean::test_kmeans2_init[array_api_strict] - ValueError: Array indexing is only allowed when array to be indexed and all indexing arrays are on the same device.
FAILED scipy/cluster/tests/test_vq.py::TestKMean::test_kmeans2_kpp_low_dim[array_api_strict] - ValueError: Array indexing is only allowed when array to be indexed and all indexing arrays are on the same device.
FAILED scipy/cluster/tests/test_vq.py::TestKMean::test_kmeans2_kpp_high_dim[array_api_strict] - ValueError: Array indexing is only allowed when array to be indexed and all indexing arrays are on the same device.
FAILED scipy/cluster/tests/test_vq.py::TestKMean::test_kmeans_and_kmeans2_random_seed[array_api_strict] - ValueError: Array indexing is only allowed when array to be indexed and all indexing arrays are on the same device.
```

CC @lucyleeow

lucyleeow:
closes #144

Fixes bug with indexing with integers introduced in https://github.com/data-apis/array-api-strict/pull/139

### DIFF
--- a/array_api_strict/tests/test_array_object.py
+++ b/array_api_strict/tests/test_array_object.py
@@ -106,7 +106,6 @@ def test_validate_index():
     "integer_index",
     [
         1,
-        np.bool(1),
         np.int8(0),
         np.uint8(0),
         np.int16(0),
--- a/array_api_strict/tests/test_array_object.py
+++ b/array_api_strict/tests/test_array_object.py
@@ -102,6 +102,31 @@ def test_validate_index():
 
 
 @pytest.mark.parametrize("device", [None, "CPU_DEVICE", "device1", "device2"])
+@pytest.mark.parametrize(
+    "integer_index",
+    [
+        1,
+        np.bool(1),
+        np.int8(0),
+        np.uint8(0),
+        np.int16(0),
+        np.uint16(0),
+        np.int32(0),
+        np.uint32(0),
+        np.int64(0),
+        np.uint64(0),
+        2,
+    ],
+)
+def test_indexing_ints(integer_index, device):
+    # Ensure indexing with different integer types works on all Devices.
+    device = None if device is None else Device(device)
+
+    a = arange(5, device=device)
+    a[integer_index]
+
+
+@pytest.mark.parametrize("device", [None, "CPU_DEVICE", "device1", "device2"])
 def test_indexing_arrays(device):
     # indexing with 1D integer arrays and mixes of integers and 1D integer are allowed
     device = None if device is None else Device(device)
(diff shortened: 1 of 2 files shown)


array_api_strict/tests/test_array_object.py
```diff
@@ -101,6 +101,30 @@ def test_validate_index():
...
+        1,
+        np.bool(1),
+        np.int8(0),
+        np.uint8(0),
+        np.int16(0),
+        np.uint16(0),
+        np.int32(0),
+        np.uint32(0),
+        np.int64(0),
+        np.uint64(0),
```
lucyleeow, line 117: Possibly overdid it, this may not be necessary ?
Also is there any integer index types i missed?

lucyleeow: nb removed `np.bool(1)` (which has a `__index__` method) as this gives warning:
"FutureWarning: In the future `np.bool` will be defined as the corresponding NumPy scalar."

array_api_strict/tests/test_array_object.py
```diff
@@ -101,6 +101,29 @@ def test_validate_index():
...
+    ],
+)
+def test_indexing_ints(integer_index, device):
+    # Ensure indexing with different integer types works on all Devices.
+    device = None if device is None else Device(device)
+
+    a = arange(5, device=device)
+    a[integer_index]
```
ev-br, line 124: Let's try to be airtight, esp given that the original issue was reported to occur for `a[(integer_index,)]`:

```suggestion
    assert a[integer_index] == a[0]
    assert a[(integer_index,)] == a[0]
```

ev-br (commented):
Agreed that `np.bool` should not be there (in the Array API land, bools are distinct).

THere's a small suggestion below. Additionally, since the OP quoted a passage about `__index__`, can we also add

```
class I:
    def __init__(self, x):
         self.x = x
    def __index__(self):
        return self.x

assert a[I(0)] == a[0]

```

array_api_strict/tests/test_array_object.py
lucyleeow: :facepalm: of course!

lucyleeow:
@ev-br thanks! Done!

ev-br:
LGTM and merged. Thanks @lucyleeow

Issue #151, opened 2025-05-15
Test failure with Python 3.14 beta 1


frenzymadness:
According to this record in the changelog of Python 3.14 beta 1:

> [gh-127682](https://github.com/python/cpython/issues/127682): No longer call __iter__ twice when creating and executing a generator expression. Creating a generator expression from a non-interable will raise only when the generator expression is executed. This brings the behavior of generator expressions in line with other generators.

the expected exception in:

https://github.com/data-apis/array-api-strict/blob/e25b9281fd9b94403ffda337d3d0240bb155f530/array_api_strict/tests/test_array_object.py#L640

is not gonna be raised until the generator returned from the `__iter__` method of `Array` object is accessed. Adding `next(…)` around the `iter(…)` call makes the test compatible with both 3.14b1 and older Pythons.

I'm testing this with older version of the package so I'm sorry if it's no longer relevant. I see the `__iter__` code is slightly different in the latest version.

ev-br:
Interesting. Would you be interested in checking if the problem also exists on main, and sending a PR with a fix if it does?

frenzymadness:
It seems the problem is in the main branch as well:

```
<mock-chroot> sh-5.2# grep -A 15 "def __iter__" array_api_strict/_array_object.py 
    def __iter__(self) -> Iterator[Array]:
        """
        Performs the operation __iter__.
        """
        # Manually disable iteration on higher dimensional arrays, since
        # __getitem__ raises IndexError on things like ones((3, 3))[0], which
        # causes list(ones((3, 3))) to give [].
        if self.ndim > 1:
            raise TypeError("array iteration is not allowed in array-api-strict")
        # Allow iteration for 1-D arrays. The array API doesn't strictly
        # define __iter__, but it doesn't disallow it. The default Python
        # behavior is to implement iter as a[0], a[1], ... when __getitem__ is
        # implemented, which implies iteration on 1-D arrays.
        return (Array._new(i, device=self.device) for i in self._array)

    def __le__(self, other: Array | float, /) -> Array:


<mock-chroot> sh-5.2# python3
Python 3.14.0b1 (main, May  7 2025, 00:00:00) [GCC 15.1.1 20250425 (Red Hat 15.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from array_api_strict import asarray
>>> iter(asarray(3))
<generator object Array.__iter__.<locals>.<genexpr> at 0x7ff2165464d0>
[...]

PR #154, opened 2025-05-16
Fix test_iter with Python 3.14 beta 1

frenzymadness:
In Python 3.14 beta 1, generator expression returned by the __iter__ method needs to be executed to throw the expected TypeError.

Fixes: https://github.com/data-apis/array-api-strict/issues/151


ev-br:
Okay, this only changes `array-api-strict` internal tests, it is reported to fix tests under python 3.14, and the CI confirms it does not affect previous CI versions. It thus looks like a no-brainer to me. Thanks @frenzymadness , it's great to see python 3.14 compliance, even if it's only in an rc phase!

PR #158, opened 2025-06-16
CI: use 500 hypothesis examples, 4 workers

ev-br:
Match the setup in array-api-compat, after https://github.com/data-apis/array-api-compat/pull/321


ev-br:
The run time is still under 5mins, with a moderate stability improvement (200-> 500 examples).

Issue #160, opened 2025-07-01
Reinstate support for Python 3.10 and 3.11


crusaderky:
With #115, you can no longer call `np.asarray(x)`, where x is a `array_api_strict.Array`, on Python 3.10 and 3.11.
This effectively means that array-api-strict is no longer fit for purpose on 3.10 and 3.11.

This minimum version bump is reflected in the release notes, but
- not in pyproject.toml
- nor in the conda-forge recipe https://github.com/conda-forge/array-api-strict-feedstock/blob/main/recipe/recipe.yaml

As a result, all projects that don't pin array-api-strict will break or have already broken.
All projects that have it as a hard dependency in their test suite (why wouldn't they? it's tiny!) break and fixing them is potentially going to require substantial effort.

I don't think this was discussed widely enough, and I, for one, am strongly convinced this was a very bad decision.

[SPEC0](https://scientific-python.org/specs/spec-0000/) dropped support for 3.10 in Q4 2024 and requires 3.11 to be supported until Q4 2025.
_Most_ projects do not aggressively enforce SPEC0. We can see some of the biggest users of array-api-strict below:

| library | Minimum Python version |
| --- | --- |
| array-api-compat | 3.10 |
| array-api-extra | 3.10 |
| scipy 1.17 | 3.11 |
| scikit-learn | 3.10 |

# Proposed course of action
## Option 1
- **Yank release 2.4;**
- Amend #115 so that it defines `__array__` on Python 3.10 and 3.11, and defines `__buffer__` instead on newer versions;
- Reinstate CI on 3.10 and 3.11;
- Issue release 2.5.

## Option 2
- **Yank release 2.4;**
- Amend pyproject.toml to prevent pypi users on Python 3.10-3.11 from upgrading;
- Issue release 2.4.1;
- Amend conda-forge recipe to prevent conda users from upgrading.

**I strongly prefer option 1**, on the basis that the users listed above would get long term harm from n.2. They won't break immediately, but will have to deal with either diverging versions of array-api-strict or changing array-api-strict to be a fully optional test dependency.

Note that Option 2 is in violation of SPEC0 until Q4 2025.

crusaderky:
CC @ev-br @lucascolley @betatim , who were involved in the original PR.

lucascolley:
I would support Option 1

betatim:
[...]

PR #161, opened 2025-07-08
BUG: officially drop Python <3.12

lucascolley:
gh-160

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


lucascolley:
could you review @crusaderky ?

lucascolley:
It was decided in the community meeting to restore support for Python 3.10+ by bringing back `__array__` for the versions that need it.

@crusaderky how should we proceed here? Do we want to make a 2.4.1 release which fixes the metadata before yanking 2.4? Or can we just jump straight to 2.5?

lucascolley:
Sounds good to me. Are you happy to make those PRs?

PR #162, opened 2025-07-11
Reinstate `__array__` on Python 3.10/3.11

### LINKED ISSUE #160: Reinstate support for Python 3.10 and 3.11

crusaderky:
- Towards #160 (need to release 2.4.1, wait some time, and yank 2.4 to close it)
- Reinstate support for Python 3.10 and 3.11
- Bump NumPy in CI from 2.2 to 2.3

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -16,6 +16,7 @@ of ndarray.
 from __future__ import annotations
 
 import operator
+import sys
 from collections.abc import Iterator
 from enum import IntEnum
 from types import EllipsisType, ModuleType
@@ -66,8 +67,6 @@ class Device:
 
 CPU_DEVICE = Device()
 ALL_DEVICES = (CPU_DEVICE, Device("device1"), Device("device2"))
-
-_default = object()
 
 
 class Array:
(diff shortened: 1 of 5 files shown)


array_api_strict/_array_object.py
```diff
@@ -149,29 +148,40 @@ def __repr__(self) -> str:
...
+    # array_api_strict.Array. So we can't get rid of __array__ entirely.
+    def __array__(
+        self, dtype: None | np.dtype[Any] = None, copy: None | bool = None
+    ) -> npt.NDArray[Any]:
+        if self._device != CPU_DEVICE:
+            # We arrive here from np.asarray() on Python >=3.12 when __buffer__ raises.
+            raise RuntimeError(
+                f"Can't convert array on the '{self._device}' device to a "
+                "NumPy array."
+            )
```
crusaderky, line 173: Previously, np.asarray() would raise `ValueError: object __array__ method not producing an array` because `__array__` subtly returned None. Changed to explicit behaviour.

lucascolley (approved):
thanks Guido, LGTM!

betatim (approved):

PR #170, opened 2025-11-17
ENH: searchorted: support `int | float` scalar `x2`

ev-br:
A matching PR for https://github.com/data-apis/array-api/pull/982 (2025.12 ehnancement) and https://github.com/data-apis/array-api-tests/pull/394 (a test).


ev-br:
tested locally with https://github.com/data-apis/array-api-tests/pull/394

PR #178, opened 2026-01-10
ENH: add linalg.eig, linalg.eigvals if Array API vesion is at least 2025.12

ev-br:
A matching test is data-apis/array-api-tests#404


ev-br:
tested locally together with data-apis/array-api-tests#404

PR #197, opened 2026-03-23
ENH: allow `copy=` in `from_dlpack`

ev-br:
cross-ref https://github.com/data-apis/array-api-tests/pull/433


ev-br:
Existing tests are passing, and so is https://github.com/data-apis/array-api-tests/pull/433 locally. Merging

PR #200, opened 2026-03-27
Upper cap setuptools_scm to <10

ev-br:
CI started failing on March 27, 2026; setuptools_scm had several 10.0.x releases on March 25-March27.
cross-ref gh-197, gh-199


ev-br:
The `build-distribution` job finishes successfully with the `setuptools_scm` upper cap. Merging straight away to unblock gh-197, gh-199.

PR #201, opened 2026-04-03
MAINT: temp skip test_from_dlpack

ev-br:
`array-api-tests` runs self-tests against a released `array-api-strict`, so skip a test until `array-api-strict==2.6` is released.


ev-br:
CI error is unrelated (test_std is flaky)

Issue #70, opened 2024-10-18
Improvements to device support


asmeurer:
https://github.com/data-apis/array-api-strict/pull/59 added support for basic devices. Some improvements that could be made:

- Factor out the device checking logic (into a decorator?)
- Add devices that don't support certain dtypes (non-support for float64/complex128 being the most common)
- Allow the user to manually create custom "devices" with certain properties (probably via the flags  API).
- Better testing for device support. This ideally should go in array-api-tests, but currently it doesn't test devices at all, so it would be useful to have some basic tests here.

ogrisel:
I tried to run the scikit-learn tests with `array-api-strict` on the non-default device and I got some failures that do not happen with PyTorch and non-default device. So I suspect there is something fishy happening with the new devices in `array-api-strict` but I have not investigated the root cause yet.

https://github.com/scikit-learn/scikit-learn/pull/30090

betatim:
I'll take a look/investigate that failure further.

lucyleeow:
Commenting here because it partially relates OP but directly related to @ogrisel 's comment https://github.com/data-apis/array-api-strict/issues/70#issuecomment-2426065761

Looking at: https://data-apis.org/array-api/latest/design_topics/device_support.html#device-support , the spec is very light in this area. It only specifies support for local control over data allocation target device, and syntax for device assignment. I agree that it makes sense that strict does not allow mixing of devices, but there are 2 different options here:

1. no mixing of devices
2. only allow the default device (note NumPy itself only allows a default device) 

Current behaviour seems a bit conflicting (?), this works:
```python
from array_api_strict import asarray, Device

asarray([1,2,3], device=Device('device1'))
```
this fails:
```python
asarray([asarray([1,2], device=Device('device1')), asarray([1,2], device=Device('device1'))])
```

due to this check:

https://github.com/data-apis/array-api-strict/blob/a8f567a184ceb2f94c6160695b63dd9c5f2454d7/array_api_strict/_array_object.py#L166-L167

this seems to be the only place where such a check is happening. And it's probably the cause of the behaviour noticed by @ogrisel where some tests on non-default devices are failing on strict but not with Torch and non-default device.

@ev-br

ev-br:
First and foremost, working on device support is very welcome @lucyleeow !

Two quick notes:

[...]

PR #206, opened 2026-04-24
ENH: add a float32-only device

### LINKED ISSUE #38: Allow changing the default dtypes
See https://data-apis.org/array-api/latest/API_specification/data_types.html#default-data-types and https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.default_dtypes.html#array_api.info.default_dtypes. 

We should add flags to the `set_array_api_strict_flags` to configure these away from the NumPy defaults. 

One concern here is that some instances of moving from float64 to float32, we might have to just downcast the result from NumPy, meaning the computation will still happen in float64, producing a result that could be different from a library that actually does everything in float32. This should likely be worked around wherever possible by downcasting the input before computing rather than the output.

### LINKED ISSUE #64: Make one of the devices have a lower maximum float precision
As a follow up to #56: it would be useful to make one of the "devices" have a lower maximum float precision than the rest. To mirror what, for example, PyTorch's MPS device does (this is the "Apple silicon" GPU in modern MacBooks).

I think having a "lower maximum precision" device would be the last piece we'd need in scikit-learn to be able to find all "user errors" when writing array API code. This way people writing array API code don't need to have a GPU available to find and debug the errors you get when using a device that isn't a CPU.

What do people think?

cc @ogrisel

ev-br:
Add a new device which only supports single precision floats and does not support double precision floats.

The new device mimics torch "mps" device in that it does not have f64 but supports int64---unlike JAX which either has both or none. 

- Make inspection APIs (`dtypes`, `default_dtypes`) device-aware;
- Make creation functions (`ones`, `empty` etc) use the device-specific default dtype when given `dtype=None, device=f32_device`

TODO:
- [x] Finish adding device-specific default dtypes  to all creation functions
- [x] device-specific default dtypes for `fft.{fftfreq, rfftfreq}`
- [x] Audit the codebase for missing `device=` arguments in internal constructions

TBD:
- [x] re-purpose an existing `device2` or add a new `F32_device` (if so, bikeshed the name)
- [x] standardize the behavior for incompatible arguments: `dtype=float64, device=f32_only_device` should raise? torch "mps" tensors raise a TypeError, follow it or mandate a ValueError?

Intends to close gh-64, 
Gives a way to close gh-38 --- if we have a f32-only device, we probably do not need a global flag
Addresses a large part of gh-70

Cross-ref the spec RFC to allow for missing dtypes , https://github.com/data-apis/array-api/issues/998 --- note that this `array-api-strict` PR can only land after the spec is updated;
Also cross-ref the test suite tracker https://github.com/data-apis/array-api-tests/issues/431: the test suite is actually fairly far along.

EDIT: https://github.com/data-apis/array-api/pull/1005 is I believe the spec update to make this spec-compliant.

### DIFF
--- a/array_api_strict/_creation_functions.py
+++ b/array_api_strict/_creation_functions.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, Literal
 
 import numpy as np
 
-from ._dtypes import DType, _all_dtypes, _np_dtype
+from ._dtypes import DType, _all_dtypes, _np_dtype, bool as xp_bool
 from ._devices import (
     CPU_DEVICE, Device, device_supports_dtype, get_default_dtypes,
     check_device as _check_device
@@ -239,12 +239,20 @@ def full(
 
     _check_device(device)
     _check_valid_dtype(dtype, device)
-    if dtype is None:
-        dtype = get_default_dtypes(device)["real floating"]
 
     if not isinstance(fill_value, bool | int | float | complex):
         msg = f"Expected Python scalar fill_value, got type {type(fill_value)}"
         raise TypeError(msg)
+
+    if dtype is None:
+        if type(fill_value) == bool:
+            dtype = xp_bool
+        else:
+            kind = {
+                int: "integral", float: "real floating", complex: "complex floating"
+            }[type(fill_value)]
+            dtype = get_default_dtypes(device)[kind]
+
     res = np.full(shape, fill_value, dtype=_np_dtype(dtype))
     if DType(res.dtype) not in _all_dtypes:
         # This will happen if the fill value is not something that NumPy
--- a/array_api_strict/_creation_functions.py
+++ b/array_api_strict/_creation_functions.py
@@ -127,6 +127,11 @@ def arange(
 
     _check_device(device)
     _check_valid_dtype(dtype, device)
+    if dtype is None:
+        if any(isinstance(x, float) for x in (start, stop, step)):
+            dtype = get_default_dtypes(device)["real floating"]
+        else:
+            dtype = get_default_dtypes(device)["integral"]
 
     return Array._new(
         np.arange(start, stop, step, dtype=_np_dtype(dtype)),
--- a/array_api_strict/_creation_functions.py
+++ b/array_api_strict/_creation_functions.py
@@ -5,7 +5,11 @@ from typing import TYPE_CHECKING, Literal
 
 import numpy as np
 
-from ._dtypes import DType, _all_dtypes, _np_dtype
+from ._dtypes import DType, _all_dtypes, _np_dtype, bool as xp_bool
+from ._devices import (
+    Device, device_supports_dtype, get_default_dtypes,
+    check_device as _check_device
+)
 from ._flags import get_array_api_strict_flags
 from ._typing import NestedSequence, SupportsBufferProtocol, SupportsDLPack
 
@@ -14,7 +18,7 @@ if TYPE_CHECKING:
     from typing_extensions import TypeIs
 
     # Circular import
-    from ._array_object import Array, Device
+    from ._array_object import Array
 
 
 class Undef(Enum):
@@ -24,10 +28,15 @@ class Undef(Enum):
 _undef = Undef.UNDEF
 
 
-def _check_valid_dtype(dtype: DType | None) -> None:
+def _check_valid_dtype(dtype: DType | None, device: Device | None = None) -> None:
     # Note: Only spelling dtypes as the dtype objects is supported.
-    if dtype not in (None,) + _all_dtypes:
-        raise ValueError(f"dtype must be one of the supported dtypes, got {dtype!r}")
+    if dtype is not None:
+        if dtype not in _all_dtypes:
+            raise ValueError(f"dtype must be one of the supported dtypes, got {dtype!r}")
+
+        if device is not None:
+            if not device_supports_dtype(device, dtype):
+                raise ValueError(f"Device {device!r} does not support dtype={dtype!r}.")
 
 
 def _supports_buffer_protocol(obj: object) -> TypeIs[SupportsBufferProtocol]:
@@ -36,18 +45,6 @@ def _supports_buffer_protocol(obj: object) -> TypeIs[SupportsBufferProtocol]:
     except TypeError:
         return False
     return True
-
-
-def _check_device(device: Device | None) -> None:
-    # _array_object imports in this file are inside the functions to avoid
-    # circular imports
-    from ._array_object import ALL_DEVICES, Device
-
-    if device is not None and not isinstance(device, Device):
-        raise ValueError(f"Unsupported device {device!r}")
-
-    if device is not None and device not in ALL_DEVICES:
-        raise ValueError(f"Unsupported device {device!r}")
 
 
 def asarray(
@@ -65,11 +62,12 @@ def asarray(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
+    _check_device(device)
+    _check_valid_dtype(dtype, device)
     _np_dtype = None
     if dtype is not None:
         _np_dtype = dtype._np_dtype
-    _check_device(device)
+
     if isinstance(obj, Array) and device is None:
         device = obj.device
 
@@ -108,6 +106,27 @@ def asarray(
         raise OverflowError("Integer out of bounds for array dtypes")
 
     res = np.array(obj, dtype=_np_dtype, copy=copy)
+
+    # numpy default dtype may differ; if so, adjust the dtype
+    if dtype is None and device is not None:
+        res_dtype = DType(res.dtype)
+        if not device_supports_dtype(device, res_dtype):
+            # find out the default dtype for the device
+            from ._data_type_functions import isdtype
+            if isdtype(res_dtype, "bool"):
+                targ_dtype = DType("bool")
+            elif isdtype(res_dtype, "integral"):
+                targ_dtype = get_default_dtypes(device)["integral"]
+            elif isdtype(res_dtype, "real floating"):
+                targ_dtype = get_default_dtypes(device)["real floating"]
+            elif isdtype(res_dtype, "complex floating"):
+                targ_dtype = get_default_dtypes(device)["complex floating"]
+            else:
+                raise ValueError(f"{res_dtype = } not understood.")
+            del isdtype
+
+            res = res.astype(targ_dtype._np_dtype)
+
     return Array._new(res, device=device)
 
 
@@ -127,8 +146,13 @@ def arange(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
+    _check_valid_dtype(dtype, device)
+    if dtype is None:
+        if any(isinstance(x, float) for x in (start, stop, step)):
+            dtype = get_default_dtypes(device)["real floating"]
+        else:
+            dtype = get_default_dtypes(device)["integral"]
 
     return Array._new(
         np.arange(start, stop, step, dtype=_np_dtype(dtype)),
@@ -149,8 +173,10 @@ def empty(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
+    _check_valid_dtype(dtype, device)
+    if dtype is None:
+        dtype = get_default_dtypes(device)["real floating"]
 
     return Array._new(np.empty(shape, dtype=_np_dtype(dtype)), device=device)
 
@@ -165,10 +191,12 @@ def empty_like(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
     if device is None:
         device = x.device
+    if dtype is None:
+        dtype = x.dtype
+    _check_valid_dtype(dtype, device)
 
     return Array._new(np.empty_like(x._array, dtype=_np_dtype(dtype)), device=device)
 
@@ -189,8 +217,10 @@ def eye(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
+    _check_valid_dtype(dtype, device)
+    if dtype is None:
+        dtype = get_default_dtypes(device)["real floating"]
 
     return Array._new(
         np.eye(n_rows, M=n_cols, k=k, dtype=_np_dtype(dtype)), device=device
@@ -242,12 +272,22 @@ def full(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
+    _check_valid_dtype(dtype, device)
 
     if not isinstance(fill_value, bool | int | float | complex):
         msg = f"Expected Python scalar fill_value, got type {type(fill_value)}"
         raise TypeError(msg)
+
+    if dtype is None:
+        if type(fill_value) == bool:
+            dtype = xp_bool
+        else:
+            kind = {
+                int: "integral", float: "real floating", complex: "complex floating"
+            }[type(fill_value)]
+            dtype = get_default_dtypes(device)[kind]
+
     res = np.full(shape, fill_value, dtype=_np_dtype(dtype))
     if DType(res.dtype) not in _all_dtypes:
         # This will happen if the fill value is not something that NumPy
@@ -271,10 +311,12 @@ def full_like(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
     if device is None:
         device = x.device
+    if dtype is None:
+        dtype = x.dtype
+    _check_valid_dtype(dtype, device)
 
     if not isinstance(fill_value, bool | int | float | complex):
         msg = f"Expected Python scalar fill_value, got type {type(fill_value)}"
@@ -305,8 +347,13 @@ def linspace(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
+    _check_valid_dtype(dtype, device)
+    if dtype is None:
+        if isinstance(start, complex) or isinstance(stop, complex):
+            dtype = get_default_dtypes(device)["complex floating"]
+        else:
+            dtype = get_default_dtypes(device)["real floating"]
 
     return Array._new(
         np.linspace(start, stop, num, dtype=_np_dtype(dtype), endpoint=endpoint),
@@ -358,8 +405,10 @@ def ones(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
+    _check_valid_dtype(dtype, device)
+    if dtype is None:
+        dtype = get_default_dtypes(device)["real floating"]
 
     return Array._new(np.ones(shape, dtype=_np_dtype(dtype)), device=device)
 
@@ -374,10 +423,12 @@ def ones_like(
     """
     from ._array_object import Array
 
-    _check_valid_dtype(dtype)
     _check_device(device)
     if device is None:
         device = x.device
+    if dtype is None:
+        dtype = x.dtype
+    _check_valid_dtype(dtype, device)
 
     return Array._new(np.ones_like(x._array, dtype=_np_dtype(dtype)), device=device)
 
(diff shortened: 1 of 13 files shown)


array_api_strict/_info.py
```diff
@@ -54,78 +49,21 @@ def dtypes(
...
-                "uint64": dt.uint64,
-                "float32": dt.float32,
-                "float64": dt.float64,
-                "complex64": dt.complex64,
-                "complex128": dt.complex128,
-            }
-        if isinstance(kind, tuple):
+        if device is None:
+            device = CPU_DEVICE
+        if isinstance(kind, type(None) | str):
```
betatim, line 54: What does this check? I was expecting to read `isinstance(kind, (type(None), str))` instead of seeing a `|`. For my education, what is the difference/when would I use one and when the other?

betatim:
I had a quick look at the diff, I think this is going in a direction I like.

One thought to which I have no good answer: would it make sense to repurpose `device2` as the one that doesn't support `float64` instead of adding "yet another device"? A reason to reuse it is to avoid having many devices, a reason to have a new device is "explicit is better than implicit" (and it is basically free to add devices).

Interested what other think about this.

array_api_strict/_info.py
ev-br: It's a union type, https://peps.python.org/pep-0604/. AFAIU the two `isinstance` forms are exactly equivalent.

FWIW, I also expect(ed) a tuple of types but apparently multiple linters pick on this in their "modernization" recommendations, similar to `%` string formatting. And some scipy reviewers felt strongly enough --- the sum total effect is apparently that my finger memory changed from `(int, float)` to `int | float`.

betatim: Every day is a school day! Thanks for the explanation

betatim: Shouldn't it be `isinstance(kind, None | str)` instead of using `type(None)`? In the PEP that form appears a few times. Python might slowly be getting too complicated for my brain

ev-br: Experimentally, this form does not seem to work:

```
In [8]: isinstance(None, None)
Cell In[8], line 1
----> 1 isinstance(None, None)

TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union

In [9]: isinstance(None, type(None))
Out[9]: True
```

Have to admit I only got the PEP reference from a google search, and I'm not sure of exact details of whether instances should or should not be allowed as the 2nd `isinstance` argument, or whether there were post-pep decisions, or if different python versions do different things. 

Here `type(None)` is what just worked so that I did not have to stop looking for alternatives. Apparently, there's `from types import NoneType` which I just now had to look up where it is defined (and am pretty sure will need to look it up again next time). If `type(None)` rubs the reader the wrong way, we can switch to that of course. 
So yeah, python is getting too complicated for my brain, too :-).

ev-br: Ah, here's the thing: 

> This PEP is a historical document. The up-to-date, canonical documentation can now be found at [Union Type](https://docs.python.org/3/library/stdtypes.html#types-union).

Looking at the Union Type reference,

> `int | str == typing.Union[int, str]`

and

> Optional types can be spelled as a union with None:
> `str | None == typing.Optional[str]`


So `int | None` and `int | type(None)` _are_ different. I'm sure there's a convincing argument (~100 comments somewhere) that `isinstance` receives a `Union` not `Optional`. Sigh.

betatim: Thanks for doing the digging. Naively `isinstance` should not accept instances, only classes/types. So `type(None)` makes sense and the "bare" `None` is somehow weird.

Hopefully LLMs can keep up with all this :D

ogrisel:
I am fine with repurposing `device2`.

ogrisel:
I think it would make sense for the float32 only device to raise an exception on `__setitem__` operations that trigger an implicit data transfer.

Note that this is not the case for PyTorch on the MPS devices for some reason:

```python
>>> import torch
>>> a = torch.ones(10, device="mps")
>>> b = 2 * torch.ones(10)  # CPU
>>> a + b
  Cell In[9], line 1
    a + b
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
>>> b[:5] = a[:5]  # implicit MPS to CPU transfer does not raise!
>>> b
tensor([1., 1., 1., 1., 1., 2., 2., 2., 2., 2.])
```

I would have expected the last statement to also raise `RuntimeError` for consistency. I think array-api-strict should raise in this case.

EDIT: actually, this is not related to dtypes. Maybe I should open a dedicated issue for this.

EDIT2: done: https://github.com/data-apis/array-api-strict/issues/207

ev-br:
Okay, this is ready from my side. 

Re yet another device vs repurpose device2, the latter is more churn; the former works today and a new device is free, so am opting to adding a new device unless someone has a strong opinion to the contrary.

ev-br:
Rebased on top of gh-212.
@betatim could I tempt you to take a look at this PR if you've the time?

rgommers:
This indeed seems useful, thanks for pushing this forward! 

> The new device mimics torch "mps" device in that it does not have f64 but supports int64---unlike JAX which either has both or none.

Also MLX on GPU: https://ml-explore.github.io/mlx/build/html/python/data_types.html

array_api_strict/_devices.py
```diff
@@ -0,0 +1,132 @@
...
+from enum import IntEnum
+
+from ._dtypes import (
+    DType, float32, float64, complex64, complex128, int64,
+    _all_dtypes, _boolean_dtypes, _signed_integer_dtypes,
+    _unsigned_integer_dtypes, _integer_dtypes, _real_floating_dtypes,
+    _complex_floating_dtypes, _numeric_dtypes
+)
+
+_ALL_DEVICE_NAMES = ("CPU_DEVICE", "device1", "device2", "F32_device")
```
betatim: How about spelling out the name of the device a bit more?

My first thought was "float32_only_device" but I think that is confusing because you might think it means the only type available is `float32`. So maybe `no_float64_device` is more obvious. I do wonder if we could call it just `no_float64`, the device postfix feels a bit superfluous, but it does make the string meaningful by itself.

```suggestion
_ALL_DEVICE_NAMES = ("CPU_DEVICE", "device1", "device2", "no_float64_device")
```

array_api_strict/_devices.py
```diff
@@ -0,0 +1,132 @@
...
+    else:
+        return {
+            "real floating": float64,
+            "complex floating": complex128,
+            "integral": int64,
+            "indexing": int64,
+        }
+
+
+def device_supports_dtype(device: Device | None, dtype: DType |None) -> bool:
```
betatim, line 105: I nearly suggested calling this `device_supports_float_dtype` but I think this function works correctly with all dtypes and all devices. Maybe we can add a comment to help people from the future. Something like "# F32_Device supports all dtypes except float64 and complex128"?

array_api_strict/_devices.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Device):
+            return False
+        return self._device == other._device
+
+    def __hash__(self) -> int:
+        return hash(("Device", self._device))
+
+    def _supported_dtypes(self) -> list[DType]:
```
betatim: Before reading the code my expectation was that we'd have the supported dtypes defined in the device's class. So something like this method. but then we'd have to make subclasses, and such.

I think I prefer having the function based approach via `get_default_dtypes`, `device_supports_dtype` and such. So I think we can remove this method.

array_api_strict/_devices.py
```diff
@@ -0,0 +1,132 @@
...
+    if device == _F32_DEVICE:
+        return dtype not in (float64, complex128)
+
+    # All other devices support all dtypes
+    return True
+
+
+def _map_supported(dtypes: list[DType], device: Device) -> dict[str, DType]:
+    return {
+        dt._canonic_name: dt
```
betatim:
```suggestion
        dt._canonical_name: dt
```
I find non abbreviated names easier to read and think about. It is a minimal cost for the creator of the code, but a huge benefit for the many people who have to read it

array_api_strict/_creation_functions.py
```diff
@@ -108,6 +106,27 @@ def asarray(
...
+                targ_dtype = DType("bool")
+            elif isdtype(res_dtype, "integral"):
+                targ_dtype = get_default_dtypes(device)["integral"]
+            elif isdtype(res_dtype, "real floating"):
+                targ_dtype = get_default_dtypes(device)["real floating"]
+            elif isdtype(res_dtype, "complex floating"):
+                targ_dtype = get_default_dtypes(device)["complex floating"]
+            else:
+                raise ValueError(f"{res_dtype = } not understood.")
+            del isdtype
```
betatim: Why do we need this `del` ? Asking because it is quite unusual to see `del` and most of the time I see it, it is used incorrectly/doesn't do what the author thinks it does.

I can't quite work out why we use it here :-/

array_api_strict/_creation_functions.py
```diff
@@ -108,6 +106,27 @@ def asarray(
...
     res = np.array(obj, dtype=_np_dtype, copy=copy)
+
+    # numpy default dtype may differ; if so, adjust the dtype
+    if dtype is None and device is not None:
+        res_dtype = DType(res.dtype)
+        if not device_supports_dtype(device, res_dtype):
+            # find out the default dtype for the device
+            from ._data_type_functions import isdtype
+            if isdtype(res_dtype, "bool"):
+                targ_dtype = DType("bool")
```
betatim: As before, I'd prefer verbose names: `target_dtype` instead of `targ_dtype`

array_api_strict/_creation_functions.py
```diff
@@ -108,6 +106,27 @@ def asarray(
         raise OverflowError("Integer out of bounds for array dtypes")
 
     res = np.array(obj, dtype=_np_dtype, copy=copy)
+
+    # numpy default dtype may differ; if so, adjust the dtype
+    if dtype is None and device is not None:
+        res_dtype = DType(res.dtype)
+        if not device_supports_dtype(device, res_dtype):
+            # find out the default dtype for the device
```
betatim: I think this does the right thing, but I was surprised by it. I was expecting the code to find the default dtype for the device, given that no dtype was passed to `asarray`, and then create the Numpy array with that information. Instead we first create the Numpy array, then check if the dtype it selected is supported by the device. This feels "backwards". I think most of that is because in my head I'm thinking "whether a device is supported or not, and if it is the default device or not are two different questions". As in, on some fictitious device the default dtype chosen by Numpy is supported, but not the default. Then this code would no longer do the right thing.

I think the reason the code is the way it is, is that without the help of `np.array` we'd have to work out which group of dtypes the input belongs to. For example in the case of `array_api_strict.asarray([1,2,3.141], device=foobar)` we'd have to determine that we need the floating point dtypes and for `array_api_strict.asarray([1,2,3], device=foobar)` we need the integer dtypes.

Maybe we can get the best of both worlds by keeping the current code and adding a check that the selected `res_dtype` is a allowed default dtype for `device`.

I think the below comment represents the current thinking and explains a bit what is happening here. I'd probably add this no matter what (unless we completely rewrite all this):
```suggestion
            # The dtype selected by Numpy might not be the default dtype
            # on this device. If the dtype is not supported by the device we
            # try to find one that is, aka the default dtype of this device.
```

array_api_strict/_devices.py
```diff
@@ -0,0 +1,132 @@
...
+    if device is not None and not isinstance(device, Device):
+        raise ValueError(f"Unsupported device {device!r}")
+
+    if device is not None and device not in ALL_DEVICES:
+        raise ValueError(f"Unsupported device {device!r}")
+
+
+# Helpers for device-specific dtype support
+
+def get_default_dtypes(device: Device | None = None) -> dict[str, Device]:
```
betatim: AI tells me: get_default_dtypes return type annotation is wrong. It is annotated -> `dict[str, Device]` but returns `dict[str, DType]`. Cosmetic, but worth fixing since the module is otherwise typed.

I'm not too bothered by static type annotations and love to hate them, but I think if AI is right we should make this one correct :D

array_api_strict/_devices.py
```diff
@@ -0,0 +1,132 @@
...
+
+def _normalize_dl_device(device_type: IntEnum | int, device_id: int) -> tuple[int, int]:
+    return (int(device_type), device_id)
+
+
+def _device_from_dlpack_device(
+    device_type: IntEnum | int, device_id: int
+) -> Device:
+    return _DLPACK_DEVICE_TO_LOGICAL.get(
+        _normalize_dl_device(device_type, device_id), CPU_DEVICE
```
betatim: This is an existing choice, so maybe a bit off topic: should we start raising if the combined device type and ID can't be found? With the current code a metal device with a device ID != 0 would magically become a CPU device. I'm not sure how you'd get the situation of a device != 0, but that just reinforces my thinking that we should make this an error. Because the likely case for this happening is if someone made a mistake somewhere.

array_api_strict/tests/test_creation_functions.py
```diff
@@ -221,6 +224,138 @@ def test_meshgrid_dtype_errors():
...
+
+def _full(a, *args, **kwds):
+    return full(a, fill_value=42.0, *args, **kwds)
+
+
+def _full_like(a, *args, **kwds):
+    return full_like(a, fill_value=42.0, *args, **kwds)
+
+
+class TestDefaultDType:
```
betatim, line 236: Did you pick the class based testing approach on purpose? I often have AI do this, even if all of the rest of the project uses test functions. Unless there is a good reason for this I'd vote for keeping it consistent and using test functions. Not classes

betatim (commented):
I like having this device! Here some review comments.

We need some kind of nice name for this device

array_api_strict/tests/test_creation_functions.py
ev-br: It's just my default mode (that it matches what AI does is by accident, I swear I'm made of meat and bone :-)) to group related tests into a class, since I find it easier to navigate a test module with a dozen classes with a dozen test functions each, than a flat list of 144 test functions. Here for example, `TestDefaultDType::test_eye` clearly tests the `eye` function for its default dtype behavior, as opposed to a module-level `test_eye` which tests the `eye` function yes, but I need to read the test to find out which behaviors of this function are being tested.

So whether it's on purpose is yes and no: no I did not give it much thought, yes I've a mild preference towards this style; the preference is mild indeed, so if this irks the reviewer, I'm happy to remove the test class. Yay/Nay?

array_api_strict/_devices.py
ev-br: Done, thanks!

array_api_strict/_devices.py
ev-br: Thanks! Since the vast majority of usage is `xp.Device("no_floa64")` with the word Device already used, I went with no `_device" postfix in the updated PR. 
If `xp.Device("no_float64_device")` reads better to you @betatim , I'll change to that.

[a totally unneccessary explanation: English is a second language for me, and my first language has a rather strong preference for not repeating words even at a cost of breaking the word order in a sentence, thus the choice above; all that said, I'm not going to argue the point if you prefer `Device("no_float64_device")`]

array_api_strict/_devices.py
ev-br: Thanks yes, it's a remnant of previous iterations and it's gone in the updated PR.

array_api_strict/_devices.py
ev-br: Done thanks.
TIL that _canonic_ is not an exact synonym for _canonical_. Thanks Tim and Merriam-Webster dictionary :-).

array_api_strict/_creation_functions.py
ev-br: Huh yes, it was there to counter a local import but it's the end of the function, and the local name is going to be garbage collected when the function exits, which is the next line. `del` is gone in the updated PR.

array_api_strict/_creation_functions.py
ev-br: Agreed that the code is a bit awkward. And indeed I'd rather lean on numpy and correct afterwards and avoid scanning the inputs to find out the dtype from scratch which can be bit of a rabbit hole, too. 

> As in, on some fictitious device the default dtype chosen by Numpy is supported, but not the default. Then this code would no longer do the right thing.

So that would be a pytorch "cpu" device with the default pytorch settings, which supports float64 but defaults to float32. That said, I don't know if we need to go all the way here? 

For now, I kept the adjust-the-dtype dance and added your explanatory comment, hoping that that'd be a reasonable compromise?

array_api_strict/_devices.py
ev-br: Agreed, done.

ev-br:
Thanks for the review @betatim ! I pushed an update which addressed most of your comments. Also took liberty to resolve those GH comments where an update seemed to just go with your suggestions, so that they don't distract from the few comments which are less clear-cut and where I'd be looking for more input.

array_api_strict/_devices.py
betatim: I prefer `"no_float64"`, so happy to go with it if you also prefer it.

array_api_strict/_devices.py
betatim: TIL that canonic is a word! I thought you were using it as an abbreviation for canonical.

array_api_strict/_creation_functions.py
betatim: > So that would be a pytorch "cpu" device with the default pytorch settings, which supports float64 but defaults to float32. That said, I don't know if we need to go all the way here?

Ah yes! I didn't remember that in pytorch we have an example of this. Maybe worth including in the comment as "example of something that won't work with the code as it is"?

Like you said getting in the business of determining the "dtype group" is a rabbit hole.

Reading this again now, can we remove the `if not device_supports_dtype(device, res_dtype):` statement and perform the code block that is in the `if` statement unconditionally? I think that would mean that we use numpy to find the "dtype group" (int, float, bool, ...) and then find the default dtype for the device for that group and use that. Most of the time that is probably a no-op but sometimes (e.g. pytorch cpu style device) it will change the dtype.

array_api_strict/tests/test_creation_functions.py
betatim: > It's just my default mode (that it matches what AI does is by accident, I swear I'm made of meat and bone :-))

So you are the reason AI learnt this pattern :D :D 

My reasoning is mostly that consistency is important. While adding "one more" variant is not the problem, following that logic we will quickly end up with 17 variants. At which point it is a problem (hard to read and follow and understand why there are 17 styles).  The only way I know to not get to 17 styles is to say no to every n+1 occasion.

The way I'd solve the grouping/naming/navigation issue (which is real) would be to create `tests/creation_functions/test_basics.py` and `tests/creation_functions/test_no_float64_device.py` - a subdirectory with different files for different areas of testing. But that is also a bit of restructuring of the code, which you could argue is adding a new style 🤷 

TL;DR: do as you think is best

ev-br: > So you are the reason AI learnt this pattern :D :D

Sure, go blame me for the ills of AI :-))).

I think I'll seize upon your TL;DR and keep the test class in place then :-). 
But adding one more test (in  https://github.com/data-apis/array-api-strict/pull/206/commits/572c4eb680039b677e3b6e79c68ec97d54cf7503) I'm adding on the module level then.

array_api_strict/_creation_functions.py
ev-br: Okay, I expected that even this will be a rabbit hole, but actually, no it's not! Here's https://github.com/data-apis/array-api-strict/pull/206/commits/572c4eb680039b677e3b6e79c68ec97d54cf7503, which does:
- remove `device_support_dtype` from the assarray figuring out the default dtype---it's extraneous indeed!
- repurpose "device2" to 1) support both float64 and float32, and 2) default to float32
- add a test

betatim: Cool! I think we should merge this now before the scope expands too much :D

betatim (approved):
Thanks for working on this and the rounds of nice review!

LGTM

betatim:
@rgommers do you want to look at this or ponder things about it? Otherwise hit "merge" after reading this.

I will come back tomorrow or the day after to merge this if it is not yet merged.

rgommers:
> Otherwise hit "merge" after reading this.

✅ 🚀

PR #209, opened 2026-05-09
ENH: make __setitem__ raise on a device mismatch

### LINKED ISSUE #207: Raise on implicit device transfer in `__setitem__`
Consider the following two arrays:

```python
>>> a = xp.ones(10, device=xp.Device("device2"))
>>> b = 2 * xp.ones(10)
```

As expected, it's not possible to run operations that combine two such arrays because they live on different devices:

```python
>>> a + b
  Cell In[23], line 1
    a + b
  File ~/miniforge3/envs/dev/lib/python3.13/site-packages/array_api_strict/_array_object.py:545 in __add__
    self._check_type_device(other)
  File ~/miniforge3/envs/dev/lib/python3.13/site-packages/array_api_strict/_array_object.py:241 in _check_type_device
ValueError: Arrays from two different devices (array_api_strict.Device('device2') and array_api_strict.Device('CPU_DEVICE')) can not be combined.
```

However, doing assignment works without raising an exception despite the fact that this causes an implicit data transfer between devices:

```python
>>> b[:5] = a[:5]
>>> b
Array([1., 1., 1., 1., 1., 2., 2., 2., 2.,
       2.], dtype=array_api_strict.float64)
```

Shouldn't the `__setitem__` call raise in this case?

ev-br:
closes gh-207

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -960,6 +960,10 @@ class Array:
         other = value
         if isinstance(value, (bool, int, float, complex)):
             other = self._promote_scalar(value)
+        else:
+            if value.device != self.device:
+                raise ValueError(f"mismatched devices: {self.device = } != {value.device =}.")
+
         dt = _result_type(self.dtype, other.dtype)
         if dt != self.dtype:
             raise TypeError(f"mismatched dtypes: {self.dtype = } and {other.dtype = }")
--- a/array_api_strict/tests/test_array_object.py
+++ b/array_api_strict/tests/test_array_object.py
@@ -220,6 +220,14 @@ def test_setitem_invalid_promotions():
         a[0] = asarray(3.5j, dtype=complex128)
 
 
+def test_setitem_device_transfer():
+    a = arange(3)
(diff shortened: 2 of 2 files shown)


ogrisel:
LGTM: I checked that the current scikit-learn test suite passes against that branch of array-api-strict.

I also checked that changing scikit-learn code to allocate a result array with `xp.empty` without setting the expected `device=` kwarg make the tests fail early, when calling `__setitem__` time, as expected.

ev-br:
https://github.com/data-apis/array-api/pull/1005 is I believe the spec update to make this spec-compliant.

ev-br:
https://github.com/data-apis/array-api/pull/1005 has been merged, so this PR is unblocked. Formally, the spec update will only be effective in the next release, which, if it's release history is any guide, can be expected in early 2027. 

While the remit of array-api-strict is to scrupulously follow the official, published spec to the letter, delaying this PR by at least half year feels impractical and unhelpful for the ongoing adoption effort. So let's let make a pragmatic move and land this now.
Thanks Olivier for the review.

PR #212, opened 2026-06-16
Add DLPack device number to arrays

betatim:
While doing some work on scikit-learn I discovered that the device type ID for arrays that are not using the default device are still equal to the CPU device type ID. The reason I noticed is that I was exploring using the DLPack device type ID as a way to tell if something was on the CPU or "some other device" in a array namespace neutral way. Being able to use array-api-strict to test code and pretend to be on a "GPU" is useful because you can run those tests without actually having a "not CPU" device.

I think it makes sense to make the device type ID consistent with how the additional devices in array-api-strict work. They are meant to be devices from which you can't `np.asarray(my_array)` and simulate a "GPU" device that has its own memory, etc.

The result is this PR. This is how things look when you use it:
```python
import array_api_strict as xp
from array_api_strict import from_dlpack

# Each logical device maps to a distinct DLPack (device_type, device_id)
for name in ("CPU_DEVICE", "device1", "device2"):
    device = xp.Device(name) if name != "CPU_DEVICE" else xp.CPU_DEVICE
    a = xp.asarray([1, 2, 3], device=device)
    print(f"{name:12} -> {a.__dlpack_device__()}")
# CPU_DEVICE   -> (<DLDeviceType.kDLCPU: 1>, 0)
# device1      -> (<DLDeviceType.kDLCUDA: 2>, 0)
# device2      -> (<DLDeviceType.kDLCUDA: 2>, 1)

# from_dlpack preserves the logical device when device=None
a = xp.asarray([1, 2, 3], device=xp.Device("device1"))
b = from_dlpack(a)
assert b.device == a.device
assert b.__dlpack_device__() == (2, 0)  # kDLCUDA, device 0
```

I am not particularly attached to the actual implementation. It is the result of asking AI to make a sketch of how this would look and feel. I've not tried to find things that break as a result of this change beyond writing new tests and running the test suite. You could argue that this should be an issue instead of a PR, but I had the code so I thought I'd put it up so people can look at it.

What do y'all think about the idea of changing the device type ID for arrays that use one of the non-default devices?

Kinda related to #70

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -69,6 +69,35 @@ CPU_DEVICE = Device()
 ALL_DEVICES = (CPU_DEVICE, Device("device1"), Device("device2"))
 
 
+class DLDeviceType(IntEnum):
+    kDLCPU = 1
+    kDLCUDA = 2
+
+
+_DLPACK_DEVICE_FOR: Final[dict[Device, tuple[DLDeviceType, int]]] = {
+    CPU_DEVICE: (DLDeviceType.kDLCPU, 0),
+    Device("device1"): (DLDeviceType.kDLCUDA, 0),
+    Device("device2"): (DLDeviceType.kDLCUDA, 1),
+}
+
+_DLPACK_DEVICE_TO_LOGICAL: Final[dict[tuple[int, int], Device]] = {
+    (int(device_type), device_id): logical_device
+    for logical_device, (device_type, device_id) in _DLPACK_DEVICE_FOR.items()
+}
+
+
+def _normalize_dl_device(device_type: IntEnum | int, device_id: int) -> tuple[int, int]:
+    return (int(device_type), device_id)
+
+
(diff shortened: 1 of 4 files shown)


ev-br:
Makes sense!
To make sure I understand the main point: the idea is that array-api-strict emulates a box with two CUDA cards?
As opposed to, say, one CUDA card and one other incompatible device, or CPU_PINNED maybe (whatever it is, actually)?. OTOH, gh-206 probably covers an addition "mps"-like device.

> using the DLPack device type ID as a way to tell if something was on the CPU or "some other device" in a array namespace neutral way.

Maybe run this past the community meeting on Thursday? 

cc @tylerjereddy for viz

ev-br:
We briefly discussed this in the community meeting on June 25, and the reception was positive. The implementation LGTM, and the conflict with gh-206 will need to be resolved one way or another. So no reason to not merge this. Thanks @betatim !

PR #214, opened 2026-06-30
ENH: allow complex inputs in std and var

ev-br:
cross-ref data-apis/array-api#1007, data-apis/array-api-tests#437

### DIFF
--- a/array_api_strict/_statistical_functions.py
+++ b/array_api_strict/_statistical_functions.py
@@ -153,8 +153,8 @@ def std(
     keepdims: bool = False,
 ) -> Array:
     # Note: the keyword argument correction is different here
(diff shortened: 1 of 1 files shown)


array_api_strict/_statistical_functions.py
```diff
@@ -153,8 +153,8 @@ def std(
     keepdims: bool = False,
 ) -> Array:
     # Note: the keyword argument correction is different here
-    if x.dtype not in _real_floating_dtypes:
-        raise TypeError("Only real floating-point dtypes are allowed in std")
+    if x.dtype not in _floating_dtypes:
```
ev-br, line 156: Gate on the array API version?

PR #221, opened 2026-07-10
BUG: Revert dlpack device

### LINKED ISSUE #219: Remove DLPack device number from arrays (revert #212)
In #212 we changed the device ID reported by the "not CPU" devices in array-api-strict. I think we should revert that change again. For one my original use-case has probably gone away but more importantly it seems to break `torch.from_dlpack` (and maybe others).

In https://github.com/scikit-learn/scikit-learn/issues/34456 you have what happens in a scikit-learn test where we try to move an array from one namespace to another. Neither device really uses CUDA, but torch tries to use its CUDA machinery because array-api-strict device claims to be a CUDA device. I haven't tried but I assume that if I had a torch with CUDA support installed this would break at a later stage when it tries to access the actual memory.

All this makes me think that faking the device ID is a step too far.

If we decide to do this we also need to check the device ID reported by the other devices in array-api-strict (`"no_float64"`)

A related thought is: what testing could we add to the array-api-strict test suite so that we'd have caught this problem before merging the PR. Should we have a CI job that runs some tests that are about checking compatibility/interaction with a library like torch?

ev-br:
closes gh-219

reverting gh-212 is a bit manual since after landing, things moved around between `_array_object.py` and `_device.py`

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -40,10 +40,7 @@ from ._dtypes import (
     _real_to_complex_map,
     _result_type,
 )
-from ._devices import (
-    CPU_DEVICE, Device, device_supports_dtype, _normalize_dl_device, _DLPACK_DEVICE_FOR,
-    DLDeviceType
-)
+from ._devices import CPU_DEVICE, Device, device_supports_dtype
 from ._flags import get_array_api_strict_flags, set_array_api_strict_flags
 from ._typing import PyCapsule
 
@@ -614,31 +611,12 @@ class Array:
                 raise NotImplementedError("The copy argument to __dlpack__ is not yet implemented")
 
             return self._array.__dlpack__(stream=stream)
-
-        kwargs: dict[str, Any] = {'stream': stream}
-        self_dl_device = _normalize_dl_device(*_DLPACK_DEVICE_FOR[self._device])
-        cpu_dl_device = _normalize_dl_device(DLDeviceType.kDLCPU, 0)
-        numpy_dl_device: tuple[IntEnum, int] | None = None
-
-        if dl_device not in [_undef, None]:
-            requested = _normalize_dl_device(dl_device[0], dl_device[1])
-            if requested == self_dl_device:
-                pass
-            elif requested == cpu_dl_device and self_dl_device != cpu_dl_device:
-                if copy is False:
-                    raise BufferError(
-                        "Cannot export array to CPU without copying when copy=False"
-                    )
-                if copy is _undef:
-                    copy = True
-                numpy_dl_device = (DLDeviceType.kDLCPU, 0)
-            else:
-                raise BufferError("unsupported device requested")
-
-        if max_version is not _undef:
-            kwargs['max_version'] = max_version
-        if numpy_dl_device is not None:
-            kwargs['dl_device'] = numpy_dl_device
+        else:
+            kwargs = {'stream': stream}
+            if max_version is not _undef:
+                kwargs['max_version'] = max_version
+            if dl_device is not _undef:
+                kwargs['dl_device'] = dl_device
         if copy is not _undef:
             kwargs['copy'] = copy
         return self._array.__dlpack__(**kwargs)
@@ -647,7 +625,8 @@ class Array:
         """
         Performs the operation __dlpack_device__.
         """
-        return _DLPACK_DEVICE_FOR[self._device]
+        # Note: device support is required for this
+        return self._array.__dlpack_device__()
 
     def __eq__(self, other: Array | complex, /) -> Array:  # type: ignore[override]
         """
--- a/array_api_strict/_creation_functions.py
+++ b/array_api_strict/_creation_functions.py
@@ -246,7 +246,6 @@ def from_dlpack(
         if copy is not _undef:
             raise ValueError("The copy argument to from_dlpack requires at least version 2023.12 of the array API")
 
-    # Going to wait for upstream numpy support
     if device is not _undef:
         _check_device(device)
     else:
--- a/array_api_strict/_creation_functions.py
+++ b/array_api_strict/_creation_functions.py
@@ -246,15 +246,12 @@ def from_dlpack(
         if copy is not _undef:
             raise ValueError("The copy argument to from_dlpack requires at least version 2023.12 of the array API")
 
+    # Going to wait for upstream numpy support
     if device is not _undef:
         _check_device(device)
     else:
         device = None
-        if hasattr(x, "__dlpack_device__"):
-            from ._devices import _device_from_dlpack_device
 
-            dl_type, dl_id = x.__dlpack_device__()
-            device = _device_from_dlpack_device(dl_type, dl_id)
     if copy in [_undef, None]:
         # numpy 1.26 does not have the copy= arg
         return Array._new(np.from_dlpack(x), device=device)
--- a/array_api_strict/tests/test_creation_functions.py
+++ b/array_api_strict/tests/test_creation_functions.py
@@ -402,12 +402,3 @@ def test_from_dlpack_default_device():
     z = from_dlpack(np.asarray([1, 2, 3]))
     assert x.device == y.device == z.device == CPU_DEVICE
 
-
-@pytest.mark.parametrize(
-    "device",
-    [Device("device1"), Device("device2")],
-)
-def test_from_dlpack_preserves_device(device):
-    x = asarray([1, 2, 3], device=device)
-    y = from_dlpack(x)
-    assert y.device == device
--- a/array_api_strict/_devices.py
+++ b/array_api_strict/_devices.py
@@ -39,14 +39,12 @@ ALL_DEVICES = (CPU_DEVICE, Device("device1"), Device("device2"), NO_FLOAT64_DEVI
 class DLDeviceType(IntEnum):
     kDLCPU = 1
     kDLCUDA = 2
-    kDLMETAL = 8
 
 
 _DLPACK_DEVICE_FOR: Final[dict[Device, tuple[DLDeviceType, int]]] = {
     CPU_DEVICE: (DLDeviceType.kDLCPU, 0),
     Device("device1"): (DLDeviceType.kDLCUDA, 0),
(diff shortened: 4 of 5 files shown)


ev-br:
Okay, this one is ready from my side. Please do check me, since it's a bit manual and not just a straight revert.

array_api_strict/_array_object.py
```diff
@@ -614,31 +611,12 @@ def __dlpack__(
...
-                raise BufferError("unsupported device requested")
-
-        if max_version is not _undef:
-            kwargs['max_version'] = max_version
-        if numpy_dl_device is not None:
-            kwargs['dl_device'] = numpy_dl_device
+        else:
+            kwargs = {'stream': stream}
+            if max_version is not _undef:
+                 kwargs['max_version'] = max_version
```
adrian740: there is an extra space here before "kwargs['max_version'] = ..."

ev-br: thanks! fixed and force-pushed.

array_api_strict/_creation_functions.py
```diff
@@ -242,15 +242,12 @@ def from_dlpack(
         if copy is not _undef:
             raise ValueError("The copy argument to from_dlpack requires at least version 2023.12 of the array API")
 
+    # Going to wait for upstream numpy support
```
betatim: I don't understand this comment :-/

array_api_strict/tests/test_creation_functions.py
```diff
@@ -395,12 +395,3 @@ def test_from_dlpack_default_device():
     z = from_dlpack(np.asarray([1, 2, 3]))
     assert x.device == y.device == z.device == CPU_DEVICE
 
-
-@pytest.mark.parametrize(
```
betatim, line 406: We could keep this test no? It seems like it checks something that is somewhat related to the DLPack device ID but also not related.

I'd keep it (unless I'm missing something)

betatim (commented):
Two small comments.

Ideally we'd have a test that uses pytorch so that we encode that learning about DLPack device IDs into a test (as in: don't mess with them). Can we do that without having to install pytest? Maybe a test that no matter which array-api-strict device is used the device ID is always CPU?

array_api_strict/_creation_functions.py
ev-br: git blame attributes it to the very first PR from Aaron which brought the initial dlpack support. Not exactly sure what that referred to; at a guess, that was `copy=` which was not implemented in numpy at the time. Either way, since neither of us two knows what it refers to, I'll take it as a sign that the comment is no longer useful and drop it.

array_api_strict/tests/test_creation_functions.py
ev-br: An issue is that this test just did not work in 2.5? So if we're just reverting the change made in 2.6, the test is broken; if we also fix `from_dlpack`, that's not just a revert...

```
In [1]: import array_api_strict as xp

In [2]: xp.__version__
Out[2]: '2.5'

In [3]: a = xp.arange(3.0, device=xp.Device('device1'))

In [4]: xp.from_dlpack(a).device
Out[4]: array_api_strict.Device('CPU_DEVICE')
```

-----------------

An attempt at a fix:

```diff
$ git diff
diff --git a/array_api_strict/_creation_functions.py b/array_api_strict/_creation_functions.py
index 685044d..5e40771 100644
--- a/array_api_strict/_creation_functions.py
+++ b/array_api_strict/_creation_functions.py
@@ -250,6 +250,8 @@ def from_dlpack(
         _check_device(device)
     else:
         device = None
+        if hasattr(x, "device"):
+            device = x.device
 
     if copy in [_undef, None]:
         # numpy 1.26 does not have the copy= arg
```

Looks simple and correct, right?

```
array_api_strict/tests/test_creation_functions.py::test_from_dlpack_default_device FAILED                           [ 96%]
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

    def test_from_dlpack_default_device():
        x = asarray([1, 2, 3])
        y = from_dlpack(x)
        z = from_dlpack(np.asarray([1, 2, 3]))
>       assert x.device == y.device == z.device == CPU_DEVICE
E       AssertionError: assert array_api_strict.Device('CPU_DEVICE') == 'cpu'
E        +  where array_api_strict.Device('CPU_DEVICE') = Array([1, 2, 3], dtype=array_api_strict.int64).device
E        +  and   'cpu' = Array([1, 2, 3], dtype=array_api_strict.int64, device=cpu).device

```

So one cannot just reuse the device of `x`, one has to map the device of `x` onto the device of array-api-strict! And the device naming is not standardized, so pretty much the only way to map them is via DLPACK enum value, but that's what we reverting here. 

TBH I don't see a way out, do you @betatim ?

ev-br:
In view of https://github.com/data-apis/array-api-strict/pull/221#discussion_r3631240064, my plan is to wait for a couple of days to see if somebody tells me I miss something simple; and if not, then land this PR and cut a 2.6.1 release. 

And then we'll be able to continue crafting dlpack improvements.

ev-br:
In the interest of unblocking downstream work, let's land this as is, and cut a 2.6.1 bugfix release.
Thanks for the review @betatim , let's continue polishing the dlpack support in a follow-up or several.

array_api_strict/tests/test_creation_functions.py
betatim: The reason I thought that this test should "just work" is that both `x` and `y` are array-api-strict arrays. But spending two minutes staring at the newly failing test in your comment doesn't really help me understand why your proposed fix shouldn't work.

I think this means that this is more complicated than I think :D So fine to fix this in the future or never or realise that this isn't actually broken

betatim:
Thanks for moving this forward. I was out since Friday, so only just catching up with things again. Looks good to me/I support your decision

array_api_strict/tests/test_creation_functions.py
ev-br: This one is confusing yes! And I had a head start to stare at this during EuroScipy. The bottom line AFAIU is this. Currently (in 2.5 and 2.6.1):

```
In [1]: import numpy as np

In [2]: import array_api_strict as xp

In [4]: a = np.ones(3)

In [5]: b = xp.from_dlpack(a)

In [6]: b.device
Out[6]: array_api_strict.Device('CPU_DEVICE')

In [7]: a.device
Out[7]: 'cpu'
```

The failed fix from https://github.com/data-apis/array-api-strict/pull/221#discussion_r3631240064 was checking `hasattr(a, "device")` which was a Yes, and then it attached `a.device` to `b`. But `a.device` is a string `'cpu`' not an `xp.Device` instance. 

Note that it currently works mainly by chance: `xp.from_dlpack` uses the default device unless explicitly provided as a kwarg:

```
In [8]: x = xp.ones(3, device=xp.Device("device1"))

In [9]: xp.from_dlpack(x).device
Out[9]: array_api_strict.Device('CPU_DEVICE')
```

So the pickle we're in for `xp.from_dlpack(x)` is this:

- either we do `if hasattr(x, "device"): device=x.device` dance --- and then if `x` is a foreign array, we attach a foreign device to an xp-array;
- or we do what 2.5 was doing, which is `device=None` means the array-api-strict default device---and then if `x` is an xp-array on a non-default device, we effectively transfer it to the default `"CPU_DEVICE"`.

Your test was guarding for the latter, but the fix was breaking the former. 

-----------

Now, the 2.5 and 2.6.1 behavior, strictly speaking, is questionable in view of [the spec requirement](https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.from_dlpack.html) (emphasis mine)

> device on which to place the created array. If device is None and x supports DLPack, the output array **must be on the same device as `x`**. Default: None.

So which array-api-strict device is the same as NumPy's `"cpu"` device? And if `"device1"` is different, then how to tell it is different without somehow mapping the device values between different libraries?

betatim: Naively (as an unsuspecting user) the behaviour that the standard describes makes sense. It is what I'd expect to happen.

That the device changes in:
```
In [8]: x = xp.ones(3, device=xp.Device("device1"))

In [9]: xp.from_dlpack(x).device
Out[9]: array_api_strict.Device('CPU_DEVICE')
```
seems super weird as a user because we don't change array library in the process.

I think attaching the device from a different array library to the result of `from_dlpack` is the wrong thing to do. The standard says that how a device is specified is an implementation detail of the library (some use a string, some a class, etc). So attaching a pytorch device to a cupy array won't do anything useful, as none of the cupy code will understand it.

I think the question of "how do you translate a device?" is the key. Can we use the [device type and ID](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack_device__.html) of the source array to find the corresponding device in array-api-strict? 
```
In [3]: x = torch.asarray([1.,2,.3])

In [4]: x.__dlpack_device__()
Out[4]: (<DLDeviceType.kDLCPU: 1>, 0)

In [5]: x = torch.asarray([1.,2,.3], device="mps")

In [6]: x.__dlpack_device__()
Out[6]: (<DLDeviceType.kDLMetal: 8>, 0)
```

So I'd adjust the device handling in https://github.com/data-apis/array-api-strict/blob/f02a5daf11f223bedb252b8c17b607b956dee778/array_api_strict/_creation_functions.py#L234 to call `__dlpack_device__()` and see if there is a matching device in array-api-strict if the argument `device=None`. My intuition would be that `__dlpack_device__` returns `(1, 0)` for `Device("CPU_DEVICE")` and `(1, 1)` for `Device("device1")`. Or something like that. As we learnt we can't change the type because that screws up things, but I think we can use the device ID to tell these otherwise similar devices apart.

ev-br: Agreed that the current behaviour is weird! And agreed about the key question, too. Any chance you've cycles to spare to prototype a solution? You seem to have a setup for downstream testing (in scikit-learn or where was that the previous attempt wrought havoc in?).

PR #224, opened 2026-07-12
ENH: add a `no x64` device

ev-br:
Add a "no_x64" device which only supports 32-bit ints and floats.

This device mimics the JAX's default (and the name follows the  JAX `jax_enable_x64` flag name). 
As discussed and requested in the Array API meeting on Jul 09.

For convenience, this PR is on top of gh-221, so only the last commit, c180b0e2cec28dcb7253dd9d0de0a290c14ac06b,  is relevant.
Since we're getting a 2.6.1 release "quickly", and a `no_float64` device was added in 2.6, I'm thinking about sneaking this new device into 2.6.1, too. Thoughts?

### DIFF
--- a/array_api_strict/_devices.py
+++ b/array_api_strict/_devices.py
@@ -2,13 +2,13 @@ from typing import Final
 from enum import IntEnum
 
 from ._dtypes import (
-    DType, float32, float64, complex64, complex128, int64,
+    DType, float32, float64, complex64, complex128, int64, uint64, int32,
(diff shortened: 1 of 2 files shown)


ev-br:
I think I'm going to sneak this into 2.6.1, unless someone objects. Please speak up if you'd prefer to not cram this into a bugfix release --- the 2.6.1 timeline is "as soon as gh-221 is ready".

PR #233, opened 2026-07-29
Improve DLPack support for non CPU devices

betatim:
This improves the `from_dlpack` behaviour for round-tripping array-api-strict arrays and raises an error for arrays that are not on the CPU device.

This is a follow up to #219 and #221 

We use the device ID to tell the difference between the different "devices" present in array-api-strict. This seems to work/not bother libraries like torch or numpy. The device type is always CPU. In #212 we tried using different device type IDs, but that turns out not to work. I think the standard says (or suggests?) that for the CPU device the only device ID that makes sense is zero. In practice it seems every library ignores this field for CPU devices. Originally I thought we needed to use different device IDs to be able to make the following code work:
```python
x = asarray([1, 2, 3], device=device)
y = from_dlpack(x)
assert y.device == device
```
Before this PR the assert will fail.

But, now we have a `isinstance(x, Array)` branch in `from_dlpack` (line 257 of `array_api_strict/_creation_functions.py`). So we could do some special casing in this `if` statement. That way we don't have to report IDs that are not part of the standard. The downside is that `__dlpack_device__` would report the same for all the devices in array-api-strict. I can't make up my mind which option I prefer.

On `main` it is possible to "transfer" a device from `"device1"` to the CPU by using `np.from_dlpack(x_on_device1)`. This is explicitly not allowed when using `np.asarray()`. This PR adjusts the DLPack behaviour to match the `__buffer__` based behaviour.

This also installs torch for one of the CI jobs to test DLpack'ing arrays to and from a different library.

### DIFF
--- a/array_api_strict/_devices.py
+++ b/array_api_strict/_devices.py
@@ -60,21 +60,19 @@ class DLDeviceType(IntEnum):
 
 
 # All the devices of array_api_strict are fictitious and their data lives in host
-# memory, so they all report the CPU device type and are told apart by their device
-# id. Reporting anything else makes consumers such as pytorch dispatch to the
-# machinery of a device they cannot actually reach, see
+# memory, so they all report the CPU device, which is device number zero. Reporting
+# anything else makes consumers such as pytorch dispatch to the machinery of a
+# device they cannot actually reach, see
 # https://github.com/data-apis/array-api-strict/issues/219
+# The logical devices cannot be told apart through DLPack, which is fine: only the
+# CPU device can be exported at all, and from_dlpack recovers the logical device of
+# an array from this library without going through DLPack.
 _DLPACK_DEVICE_FOR: Final[dict[Device, tuple[DLDeviceType, int]]] = {
-    CPU_DEVICE: (DLDeviceType.kDLCPU, 0),
-    Device("device1"): (DLDeviceType.kDLCPU, 1),
-    Device("device2"): (DLDeviceType.kDLCPU, 2),
-    NO_FLOAT64_DEVICE: (DLDeviceType.kDLCPU, 3),
-    NO_X64_DEVICE: (DLDeviceType.kDLCPU, 4),
+    device: (DLDeviceType.kDLCPU, 0) for device in ALL_DEVICES
 }
 
 _DLPACK_DEVICE_TO_LOGICAL: Final[dict[tuple[int, int], Device]] = {
-    (int(device_type), device_id): logical_device
-    for logical_device, (device_type, device_id) in _DLPACK_DEVICE_FOR.items()
+    (int(DLDeviceType.kDLCPU), 0): CPU_DEVICE,
 }
 
 
--- a/array_api_strict/_devices.py
+++ b/array_api_strict/_devices.py
@@ -1,3 +1,4 @@
+from enum import IntEnum
 from typing import Final
 
 from ._dtypes import (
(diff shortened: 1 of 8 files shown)


betatim:
To illustrate my indecision, the second commit removes the special "use different device IDs for different logical devices". I think it is cleaner this way

array_api_strict/_devices.py
```diff
@@ -39,6 +40,64 @@ def __hash__(self) -> int:
...
+    kDLMetal = 8
+    kDLVPI = 9
+    kDLROCM = 10
+    kDLROCMHost = 11
+    kDLExtDev = 12
+    kDLCUDAManaged = 13
+    kDLOneAPI = 14
+    kDLWebGPU = 15
+    kDLHexagon = 16
+    kDLMAIA = 17
```
ev-br: Nit: The spec mandates _valid device type enum members_ without a leading `kDL` suffix: 
https://data-apis.org/array-api/draft/API_specification/generated/array_api.array.__dlpack_device__.html#dlpack-device

I wonder why and whether we'd rather change it in the Array API spec to match the DLPack spec.

array_api_strict/tests/test_array_object.py
```diff
@@ -759,6 +761,61 @@ def test_dlpack_2023_12(api_version):
...
+    # represent a GPU or other accelerator, so the consumer would end up with
+    # the data on a device the array is not logically on.
+    with pytest.raises(BufferError):
+        a.__dlpack__(dl_device=a.__dlpack_device__())
+    with pytest.raises(BufferError):
+        a.__dlpack__(dl_device=(DLDeviceType.kDLCPU, 0))
+    with pytest.raises(BufferError):
+        np.from_dlpack(a, device="cpu")
+
+    np.from_dlpack(a.to_device(CPU_DEVICE))
```
ev-br: How about 

```suggestion
    # explicitly move the array to CPU_DEVICE before handing to to DLPack
    a_np = np.from_dlpack(a.to_device(CPU_DEVICE))
    assert (a_np == a._array).all()
    assert a_np.dtype == a._array.dtype
```

or some such

array_api_strict/tests/test_creation_functions.py
```diff
@@ -413,3 +413,29 @@ def test_from_dlpack_default_device():
...
+            return (DLDeviceType.kDLCUDA, 0)
+
+        def __dlpack__(self, **kwargs):
+            return self._array.__dlpack__(**kwargs)
+
+    with pytest.raises(BufferError):
+        from_dlpack(ForeignArray())
+
+    # an explicit device is a request to transfer, so nothing has to be inferred
+    assert from_dlpack(ForeignArray(), device=CPU_DEVICE).device == CPU_DEVICE
```
ev-br, line 441: Do we want to cover `copy={True, False}` here, too?

ev-br (commented):
This looks pretty comprehensive! 
I agree with the decision to not fiddle with device IDs. I don't know if the standards mandates that CPU device ID is always 0, but even if it does not, what would `<CPU, 3>` mean anyway---a third node on a multinode HPC cluster? If anything, previous experience emulating these dlpack details showed that it can backfire in unexpected ways. So since fake device IDs are not strictly necessary, I agree with keeping it all minimal.

The comments below are all minor, really. One thing that stands out is the difference in naming of DLPack enum values between the DLPack spec and the Array API spec; maybe this is something to fix in the latter?

Adding a test run against torch is great actually.

Testing: my brain hurts when I try to trace through how this interacts with `copy={True, False, None}`. Maybe testing can be extended to cover various values of `copy=`?

And there are merge conflicts because of the ruff refactor (no comment).

array_api_strict/_devices.py
betatim: I have no opinion on the names here, but I think this being Python we should probably match the names in https://data-apis.org/array-api/draft/API_specification/generated/array_api.array.__dlpack_device__.html#dlpack-device instead of what I did (take them from the DLPack header file). The `kDLCPU` feels very "C".

I also removed a few entries from the enum, those that are not mentioned on the array API spec website. We are "the strict library" after all

betatim: Mark as resolved?

ev-br: I think it's worth a brief mention in the community meeting , if only to double-check the overwhelming support for dropping the C-style prefixes (and they do feel very C indeed).

array_api_strict/tests/test_creation_functions.py
betatim: Can we leave it for a new PR? I'm not yet sure I fully grok how it should work (`copy=`)

array_api_strict/_devices.py
betatim: I made an entry in the hackmd

betatim:
Conflicts resolved. I'd prefer to leave the `copy=` questions to a new PR, mostly because I'd have to spend some time thinking about it first

ev-br:
I ran array-api-tests on this branch with a large number of examples, and it passed. Also checked that the SciPy tests suite passes with array-api-strict from this branch. That in itself is not very surprising, since SciPy testing of devices is rather thin, but we have a confirmation that changes here don't break it.
Do you prefer to test scikit-learn before landing this or after @betatim ?

betatim:
I checked scikit-learn's tests. Of course something fails, but I suspect it might be legit bug in scikit-learn? I'll keep you posted.

ev-br:
From my side, the PR status is ready as soon as Tim says it is.

ev-br:
This PR seems to have slipped under everyone's radar. Anything to add/change here @betatim or is it good to go in from your side?

PR #234, opened 2026-07-29
MAINT: Bump `ruff` to `v0.16`

### LINKED ISSUE #228: Unpin ruff, fix warnings from 0.16
[`ruff 0.16` changed the default rule set](https://astral.sh/blog/ruff-v0.16.0#better-default-rule-set):

> **Better default rule set**
> 
> Ruff now enables 413 rules by default, up from 59 in previous versions.

This generates a wall of warnings now, and https://github.com/data-apis/array-api-strict/pull/227 pinned `ruff` to `<0.16`.

We should unpin it and investigate.

prady0t:
Closes #228

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -22,13 +22,13 @@ from enum import IntEnum
 from types import EllipsisType, ModuleType
 from typing import Any, Literal, SupportsIndex
 
+import numpy as np
+import numpy.typing as npt
+
 try:
     from typing import Self
 except ImportError:  # Python < 3.11
     from typing_extensions import Self
-
-import numpy as np
-import numpy.typing as npt
 
 from ._creation_functions import Undef, _undef, asarray
 from ._devices import CPU_DEVICE, Device, device_supports_dtype
@@ -448,9 +448,10 @@ class Array:
         else:
             ellipsis_start = None
             for pos, i in enumerate(nonexpanding_key):
-                if not (isinstance(i, (Array, np.ndarray))) and i == Ellipsis:
-                    ellipsis_start = pos
-                    break
+                if not (isinstance(i, (Array, np.ndarray))):
+                    if i == Ellipsis:
+                        ellipsis_start = pos
+                        break
             assert ellipsis_start is not None  # sanity check
             ellipsis_end = self.ndim - (n_single_axes - ellipsis_start)
             indexed_shape = (
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -202,18 +202,19 @@ class Array:
         # to promote in the spec (even if the NumPy array operator would
         # promote them).
         res_dtype = _result_type(self.dtype, other.dtype)
-        # Note: NumPy will allow in-place operators in some cases where
-        # the type promoted operator does not match the left-hand side
-        # operand. For example,
+        if op.startswith("__i"):
+            # Note: NumPy will allow in-place operators in some cases where
+            # the type promoted operator does not match the left-hand side
+            # operand. For example,
 
-        # >>> a = np.array(1, dtype=np.int8)
-        # >>> a += np.array(1, dtype=np.int16)
+            # >>> a = np.array(1, dtype=np.int8)
+            # >>> a += np.array(1, dtype=np.int16)
 
-        # The spec explicitly disallows this.
-        if op.startswith("__i") and res_dtype != self.dtype:
-            raise TypeError(
-                f"Cannot perform {op} with dtypes {self.dtype} and {other.dtype}"
-            )
+            # The spec explicitly disallows this.
+            if res_dtype != self.dtype:
+                raise TypeError(
+                    f"Cannot perform {op} with dtypes {self.dtype} and {other.dtype}"
+                )
 
         return other
 
@@ -425,17 +426,18 @@ class Array:
         n_single_axes = len(single_axes)
         if n_ellipsis > 1:
             return  # handled by ndarray
-        # Note boolean masks must be the sole index, which we check for
-        # later on.
-        elif n_ellipsis == 0 and not key_has_mask and n_single_axes < self.ndim:
-            raise IndexError(
-                f"{self.ndim=}, but the multi-axes index only specifies "
-                f"{n_single_axes} dimensions. If this was intentional, "
-                "add a trailing ellipsis (...) which expands into as many "
-                "slices (:) as necessary - this is what np.ndarray arrays "
-                "implicitly do, but such flat indexing behaviour is not "
-                "specified in the Array API."
-            )
+        elif n_ellipsis == 0:
+            # Note boolean masks must be the sole index, which we check for
+            # later on.
+            if not key_has_mask and n_single_axes < self.ndim:
+                raise IndexError(
+                    f"{self.ndim=}, but the multi-axes index only specifies "
+                    f"{n_single_axes} dimensions. If this was intentional, "
+                    "add a trailing ellipsis (...) which expands into as many "
+                    "slices (:) as necessary - this is what np.ndarray arrays "
+                    "implicitly do, but such flat indexing behaviour is not "
+                    "specified in the Array API."
+                )
 
         if (key_has_index_array and (n_ellipsis > 0 or key_has_slices or key_has_mask)):
             raise IndexError(
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -20,7 +20,12 @@ import sys
 from collections.abc import Callable, Iterator
 from enum import IntEnum
 from types import EllipsisType, ModuleType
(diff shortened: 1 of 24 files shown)


array_api_strict/_array_object.py
```diff
@@ -17,15 +17,21 @@
...
+from collections.abc import Callable, Iterator
 from enum import IntEnum
 from types import EllipsisType, ModuleType
-from typing import Any, Literal, SupportsIndex, Callable
+from typing import Any, Literal, SupportsIndex
+
+try:
+    from typing import Self
+except ImportError:  # Python < 3.11
+    from typing_extensions import Self
```
prady0t, line 31: Either this or we will have to add `typing_extensions` as a required dependency.

ev-br:
Thanks. Does this include both `--unsafe-fixes` and `--fixes`? If so, it could be helpful to split into two PRs: just `--fixes` first and unsafe ones next.

prady0t:
Has three types of fixes:
- fix
- unsafe-fixes
- rest of the failures (mentioned in issue comments) that I fixed with AI

prady0t:
I've opened a PR for safe fixes; after it's merged, I'll open another PR for unsafe fixes. Once we have both merged and main merged with this branch, this PR will be left with the rest of the changes.

array_api_strict/tests/test_searching_functions.py
```diff
@@ -11,9 +11,11 @@ def test_where_with_scalars():
...
 
     # Versions up to and including 2023.12 don't support scalar arguments
-    with ArrayAPIStrictFlags(api_version='2023.12'):
-        with pytest.raises(AttributeError, match="object has no attribute 'dtype'"):
-            xp.where(x == 1, 42, 44)
+    with (
+        ArrayAPIStrictFlags(api_version='2023.12'),
+        pytest.raises(AttributeError, match="object has no attribute 'dtype'"),
+    ):
+        xp.where(x == 1, 42, 44)
```
prady0t: We should revert these changes too.

array_api_strict/_array_object.py
```diff
@@ -202,19 +202,18 @@ def _check_allowed_dtypes(
...
+        # >>> a = np.array(1, dtype=np.int8)
+        # >>> a += np.array(1, dtype=np.int16)
 
-            # The spec explicitly disallows this.
-            if res_dtype != self.dtype:
-                raise TypeError(
-                    f"Cannot perform {op} with dtypes {self.dtype} and {other.dtype}"
-                )
+        # The spec explicitly disallows this.
+        if op.startswith("__i") and res_dtype != self.dtype:
```
prady0t: Another nested if collapse

array_api_strict/_array_object.py
```diff
@@ -427,18 +425,17 @@ def _validate_index(
...
-                    f"{self.ndim=}, but the multi-axes index only specifies "
-                    f"{n_single_axes} dimensions. If this was intentional, "
-                    "add a trailing ellipsis (...) which expands into as many "
-                    "slices (:) as necessary - this is what np.ndarray arrays "
-                    "implicitly do, but such flat indexing behaviour is not "
-                    "specified in the Array API."
-                )
+        # Note boolean masks must be the sole index, which we check for
+        # later on.
+        elif n_ellipsis == 0 and not key_has_mask and n_single_axes < self.ndim:
```
prady0t: nested if collapse

array_api_strict/_searching_functions.py
```diff
@@ -79,10 +79,9 @@ def searchsorted(
     See its docstring for more information.
     """
     flags = get_array_api_strict_flags()
-    if flags["api_version"] >= "2025.12":
-        # scalar x2 support is new in 2025.12
-        if isinstance(x2, bool | int | float | complex):
-            x2 = x1._promote_scalar(x2)
+    # scalar x2 support is new in 2025.12
+    if flags["api_version"] >= "2025.12" and isinstance(x2, bool | int | float | complex):
+        x2 = x1._promote_scalar(x2)
```
prady0t: nested if

prady0t (commented):
Some nested if and with statements can be reverted; otherwise, the changes are okay.

array_api_strict/tests/test_array_object.py
```diff
@@ -365,7 +365,7 @@ def _array_vals():
                                 _op.startswith("__i")
                                 and result_type(x.dtype, y.dtype) != x.dtype
                             ):
-                                assert_raises(TypeError, lambda: getattr(x, _op)(y))
+                                assert_raises(TypeError, lambda x=x, _op=_op, y=y: getattr(x, _op)(y))
```
ev-br: If we're touching this anyway, then how about

```
with assert_raises(TypeError):
   getattr(x, _op)(y)
```

array_api_strict/tests/test_elementwise_functions.py
```diff
@@ -292,7 +292,10 @@ def _array_vals():
 
         for s in [1, 1.0, 1j, BIG_INT, False]:
             for a in _array_vals():
-                for func1 in [lambda s: func(a, s), lambda s: func(s, a)]:
+                for func1 in [
+                    lambda s, func=func, a=a: func(a, s),
+                    lambda s, func=func, a=a: func(s, a),
+                ]:
```
ev-br: Not sure what these are TBH, could you explain?

prady0t: That's [B023](https://docs.astral.sh/ruff/rules/function-uses-loop-variable/). 
Python uses late binding for closures, hence it's worth setting initial value for the variables.

array_api_strict/tests/test_array_object.py
prady0t: Should we replace all such occurrences with a similar change?

ev-br: If you touch them anyway, then yes, I think it'd be nice.

array_api_strict/tests/test_elementwise_functions.py
prady0t: Let's ignore this if you feel it's confusing.

array_api_strict/tests/test_array_object.py
prady0t: In the latest commit, I've replaced only in those places where Ruff was complaining.

array_api_strict/_array_object.py
```diff
@@ -542,8 +541,8 @@ def __and__(self, other: Array | int, /) -> Array:
         other = self._check_allowed_dtypes(other, "integer or boolean", "__and__")
         if other is NotImplemented:
             return other
-        self, other = self._normalize_two_args(self, other)
-        res = self._array.__and__(other._array)
+        _self, other = self._normalize_two_args(self, other)
+        res = _self._array.__and__(other._array)
```
ev-br, line 545: This pattern (repeated multiple times below) looks usunsual enough, thus I looked a bit more carefully.
The whole `_normalize_two_args` dance is needed to work around 0D numpy arrays lacking functionality: `self._normalize_two_args(self, other)` return a pair of `-strict` arrays, wher each element is either the original argument, or a new `-strict` array with its internal `_array` attribute promoted from 0D to 1D. Then, next line, `{self, _self}._array.__op__(other._array)` does the `__op__` with numpy attrbutes, and the next line is `return self.__class__._new(res, device=self.device)`, so we return a new `-strict` array object.

Either way, we do operations with numpy `_array` attributes and always return a new `-strict` array object---the actual operations are done with numpy `_array` attributes, and this is where it is decided if the actual operation is in-place or not.

Before this patch, the code did potentially mutate `self`, which was harmless because we always a new object constructed in `return self.__class__._new`. 
This patch only makes the intent clearer by renaming the potentially mutated object.
Conclusion: it is a good change.

array_api_strict/_manipulation_functions.py
```diff
@@ -142,7 +142,7 @@ def roll(
     See its docstring for more information.
     """
     if not isinstance(shift, int | tuple):
-        raise ValueError(
+        raise TypeError(
```
ev-br, line 145: Strictly speaking, this is a backwards incompatible change. 
That said, TypeError makes more sense, and spot-checking several "real" frameworks shows that a TypeError is more common (exceptions: `np.roll(np.eye(3), "oops")` raises a ValueError; `jnp.roll(jnp.eye(3), None)` raise a  ValueError too --- these are edge cases though). 

The spec does not mandate a specific exception class, thus we can accept the backwards compat break and make change it to a more appropriate exception type, as done here.

ev-br (approved):
I'm not a fan of littering code with `# noqa` line noise but that's what it is, no point arguing about it.

The change to using `assert_raises` as a context manager from lambdas is definitely good.

The `self -> _self` change in array dunders looks scarier than it is, and the end result is marginally clearer.

All in all, this patch LGTM, thank you @prady0t for seeing it through!

PR #235, opened 2026-07-30
MAINT: Ruff `safe-fixes`

prady0t:
#234 #228

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -17,15 +17,16 @@ from __future__ import annotations
 
 import operator
 import sys
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from enum import IntEnum
 from types import EllipsisType, ModuleType
-from typing import Any, Literal, SupportsIndex, Callable
+from typing import Any, Literal, SupportsIndex
 
 import numpy as np
 import numpy.typing as npt
 
 from ._creation_functions import Undef, _undef, asarray
+from ._devices import CPU_DEVICE, Device, device_supports_dtype
 from ._dtypes import (
     DType,
     _all_dtypes,
@@ -40,7 +41,6 @@ from ._dtypes import (
     _real_to_complex_map,
     _result_type,
 )
-from ._devices import CPU_DEVICE, Device, device_supports_dtype
 from ._flags import get_array_api_strict_flags, set_array_api_strict_flags
 from ._typing import PyCapsule
 
@@ -64,7 +64,7 @@ class Array:
     _array: npt.NDArray[Any]
     _dtype: DType
     _device: Device
-    __slots__ = ("_array", "_dtype", "_device", "__weakref__")
+    __slots__ = ("__weakref__", "_array", "_device", "_dtype")
 
     # Use a custom constructor instead of __init__, as manually initializing
     # this class is not supported API.
--- a/array_api_strict/_linalg.py
+++ b/array_api_strict/_linalg.py
@@ -9,7 +9,7 @@ from ._array_object import Array
 from ._data_type_functions import finfo
 from ._dtypes import DType, _floating_dtypes, _numeric_dtypes, complex64, complex128
 from ._elementwise_functions import conj
-from ._flags import get_array_api_strict_flags, requires_extension, requires_api_version
+from ._flags import get_array_api_strict_flags, requires_api_version, requires_extension
 from ._manipulation_functions import reshape
 from ._statistical_functions import _np_dtype_sumprod
 
@@ -232,7 +232,7 @@ def matrix_norm(
     *,
     keepdims: bool = False,
     ord: float | Literal["fro", "nuc"] | None = "fro",
-) -> Array:  # noqa: F821
+) -> Array:
     """
     Array API compatible wrapper for :py:func:`np.linalg.norm <numpy.linalg.norm>`.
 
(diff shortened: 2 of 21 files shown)


array_api_strict/tests/test_array_object.py
```diff
@@ -1,39 +1,42 @@
...
     int8,
     int16,
     int32,
     int64,
+    uint8,
     uint64,
-    float64,
-    complex128,
+)
+from .._dtypes import (
```
ev-br: nit: can merge this import together with the one just above?

array_api_strict/tests/test_creation_functions.py
```diff
@@ -22,11 +22,12 @@
     zeros,
     zeros_like,
 )
-from .._dtypes import float32, float64, complex64, int32, int64, bool as xp_bool
-from .._array_object import Array
-from .._devices import CPU_DEVICE, ALL_DEVICES, Device
-from .._info import __array_namespace_info__
+from .._devices import ALL_DEVICES, CPU_DEVICE, Device
+from .._dtypes import bool as xp_bool
```
ev-br: Thtat's the second time `from .._dtypes import ... as ...` is separated from the same import. Is it The Right Way To Do It, according to the Supreme Lint Authority now? 
Fine by me if so, we mere mortals will adapt (and that's not that big of a deal anyway, just unexpected).

array_api_strict/tests/test_statistical_functions.py
```diff
@@ -44,9 +46,8 @@ def test_sum_prod_trace_2023_12(func_name):
 def test_mean_complex():
     a = xp.asarray([1j, 2j, 3j])
 
-    with ArrayAPIStrictFlags(api_version='2023.12'):
-        with pytest.raises(TypeError):
-            xp.mean(a)
+    with ArrayAPIStrictFlags(api_version='2023.12'), pytest.raises(TypeError):
```
ev-br: This I've to admit the original is more readable because it clearly separates a setting of flags and an assertion (which just so happens to be a context manager, too).
What would it take to convince ruff to leave this alone? `# noqa` comments are line noise, let's not add them.

array_api_strict/_array_object.py
```diff
@@ -64,7 +64,7 @@ class Array:
     _array: npt.NDArray[Any]
     _dtype: DType
     _device: Device
-    __slots__ = ("_array", "_dtype", "_device", "__weakref__")
+    __slots__ = ("__weakref__", "_array", "_device", "_dtype")
```
ev-br, line 67: Huh. I actually thought that `__slots__` are ordered and the order translates directly to the order of elements of a corresponding C struct. Apparently not, and I actually don't know why I thought this. TIL.
(this is a note-to-self, not a request to change anything)

array_api_strict/_linalg.py
```diff
@@ -232,7 +232,7 @@ def matrix_norm(
     *,
     keepdims: bool = False,
     ord: float | Literal["fro", "nuc"] | None = "fro",
-) -> Array:  # noqa: F821
```
ev-br, line 235: Oh great, thanks for this! `#noqa:` comments are line noise.

array_api_strict/_searching_functions.py
```diff
@@ -5,7 +5,11 @@
 from ._array_object import Array
 from ._dtypes import _real_numeric_dtypes, _result_type
 from ._dtypes import bool as _bool
-from ._flags import requires_api_version, requires_data_dependent_shapes, get_array_api_strict_flags
+from ._flags import (
+    get_array_api_strict_flags,
+    requires_api_version,
+    requires_data_dependent_shapes,
+)
```
ev-br, line 11: No idea why it keeps reformatting imports from a single line of 101 symbol to a column. It can't be just line length, because there are longer lines just below.
Just a note-to-self, not a request.

ev-br (commented):
LGTM modulo a question about nested context managers.
If the fix for that is onerous, fine to keep as is, too.

array_api_strict/tests/test_statistical_functions.py
prady0t: We can ignore SIM117 globally in ruff.toml

array_api_strict/tests/test_array_object.py
prady0t: Ruff does this as a default behaviour now, but we can revert this by setting `combine-as-imports = true`. See docs: https://docs.astral.sh/ruff/settings/?utm_source=chatgpt.com#lint_isort_combine-as-imports

array_api_strict/tests/test_creation_functions.py
prady0t: > Ruff does this as a default behaviour now, but we can revert this by setting combine-as-imports = true. See docs: https://docs.astral.sh/ruff/settings/?utm_source=chatgpt.com#lint_isort_combine-as-imports

Same.

array_api_strict/tests/test_array_object.py
prady0t: Fixed in the latest commit.

array_api_strict/tests/test_statistical_functions.py
prady0t: Fixed in the latest commit.

ev-br (approved):
LGTM and seems uncontrovesial, lets' get this in. I'll squash-merge so that there's a single commit to `.git-blame-ignore-revs` per https://github.com/data-apis/array-api-strict/issues/228#issuecomment-5129216100

Thanks @prady0t

PR #236, opened 2026-07-30
MAINT: Adding ruff unsafe-fixes

prady0t:
Towards #228

### DIFF
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -24,7 +24,11 @@ from typing import Any, Literal, SupportsIndex
 
 import numpy as np
 import numpy.typing as npt
-from typing_extensions import Self
+
+try:
+    from typing import Self
+except ImportError:  # Python < 3.11
+    from typing_extensions import Self
 
 from ._creation_functions import Undef, _undef, asarray
 from ._devices import CPU_DEVICE, Device, device_supports_dtype
--- a/array_api_strict/_array_object.py
+++ b/array_api_strict/_array_object.py
@@ -24,6 +24,7 @@ from typing import Any, Literal, SupportsIndex
 
 import numpy as np
 import numpy.typing as npt
+from typing_extensions import Self
 
 from ._creation_functions import Undef, _undef, asarray
 from ._devices import CPU_DEVICE, Device, device_supports_dtype
@@ -102,7 +103,7 @@ class Array:
         return obj
 
     # Prevent Array() from working
-    def __new__(cls, *args: object, **kwargs: object) -> Array:
+    def __new__(cls, *args: object, **kwargs: object) -> Self:
         raise TypeError(
             "The array_api_strict Array object should not be instantiated directly. Use an array creation function, such as asarray(), instead."
         )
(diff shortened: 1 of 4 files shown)


array_api_strict/_array_object.py
```diff
@@ -25,6 +25,11 @@
 import numpy as np
 import numpy.typing as npt
 
+try:
+    from typing import Self
+except ImportError:  # Python < 3.11
+    from typing_extensions import Self
```
prady0t, line 31: Any better way to do this?

ev-br: TBH this whole `-> Array` -to- `-> Self` thingy is a very typical example of two kinds of nonsense: one from typing and another from linting. 

This is all make-work. It's make-work to create a patch, it's make-work to review, it's make-work to even think about any of this. Why oh why does anyone spends any part of their waking time actually thinking about whether an `__iadd__` returns an `Array` or `Self`? Why does this whole contraption asks us to read about PEP673 --- okay, I googled it, I skimmed through it, fine; however long it took me, it's wasted time --- I'm ready to believe that somebody somewhere somehow cares about this and there exists at least one situation where it matters for whatever definition of matter, okay. But why on good green earth, why does it get inflicted on us? 

Anyhow, I'm done venting. 
Short answer: from my perch, I don't care, as long as this does not get in the way  (as in, I'll probably have to pip install something next time I do actual work in an environment with python==3.10, OK, I'll do it once and move on doing actual work).

ev-br (commented):
I'd like to understand why these are *unsfafe-fixes*. The changes seem to be of two groups:

- typing `-> Array` replaced by `-> Self` (ok whatever)
      This should be perfectly safe at runtime 
- nested ifs collapsed with either `and` or `isinstance(object, tuple_of_types)` 
      The former form shortcuts fine, the latter... probably shortcuts too, not sure. 
Plus benign refactors of `if-else` into a `dict.get` etc.

And if it's labeled as *unsafe*, it means I don't see something, so what is it I'm not seeing: Which of these is unsafe and how?

array_api_strict/_array_object.py
prady0t: Fair enough 😄.

prady0t:
> I'd like to understand why these are unsfafe-fixes. 

See: https://docs.astral.sh/ruff/linter/#fix-safety

I think for the first part: the conditional import is something that I added manually; originally Ruff just imported `Self` from `typing`. But for older versions, it's only available in `typing_extensions`. And the CI did fail on that.

For the second part, I'm not sure about it either.

ev-br:
Thanks for the link. My question is still why these fixes are "unsafe" in the meaning of that link: what are the changes to the runtime behaviour.

prady0t:
After a bit more digging through the docs:

- PYI034: https://docs.astral.sh/ruff/rules/non-self-return-type/

> This rule's fix is marked as unsafe as it changes the meaning of your type annotations.

- SIM102: https://docs.astral.sh/ruff/rules/collapsible-if/

> When [preview](https://docs.astral.sh/ruff/preview/) is enabled, the fix for this rule is considered as safe. When [preview](https://docs.astral.sh/ruff/preview/) is not enabled, the fix is always considered unsafe.

ev-br:
Thanks for digging through their docs! It all makes sense now.
For SIM102 could you please keep the changes but disable the rule globally? And then merge.
Reason: starting from a nested if, it will first collapse it into one, then activate the line length rule and break the line with brackets. The net effect is even funnier indentation, plus back-and-forth for a human author.

prady0t:
> The net effect is even funnier indentation, plus back-and-forth for a human author.

Totally agreed!

prady0t:
Also added some comments in `ruff.toml`

ev-br:
Thank you @prady0t , merged.

PR #240, opened 2026-08-10
TST, CI: Skip platform dependent array API failures + adding windows and macOS env in CI

### LINKED ISSUE #238: Platform-dependent test failures in `atanh` and `tanh`
In mdhaber/marray#176, I encountered the following test failures that appear to be inherited from `array-api-strict`. However, they are not encountered in `array-api-strict` CI because they all pass on Ubuntu. It appears that all fail on macOS, and the `tanh` ones fail on Windows.

```
array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j] - AssertionError: out=(nan+nanj), but should be +0 + NaN j [atanh()]
array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is 1 and imag(x_i) is +0) -> +infinity + 0j] - AssertionError: out=(inf+0.7853981633974483j), but should be +infinity + 0j [atanh()]
array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +infinity and imag(x_i) is NaN) -> +0 + NaN j] - AssertionError: out=(nan+1.5707963267948966j), but should be +0 + NaN j [atanh()]
array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and imag(x_i) is +infinity) -> \xb10 + \u03c0j/2] - AssertionError: out=(nan+1.5707963267948966j), but should be ±0 + πj/2 [atanh()]
array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j] - AssertionError: out=(nan+nanj), but should be +0 + NaN j [tanh()]
array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j] - AssertionError: out=(nan+nanj), but should be +0 + NaN j [tanh()]
```

Not sure if this requires any action, but it would be nice if they were added to the skip file.

prady0t:
Closes: #238

### DIFF
--- a/array-api-tests-macos-xfails.txt
+++ b/array-api-tests-macos-xfails.txt
@@ -1,0 +1,9 @@
+# https://github.com/numpy/numpy/issues/32234
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_operators_and_elementwise_functions.py::test_sqrt
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is 1 and imag(x_i) is +0) -> +infinity + 0j]
+array_api_tests/test_special_cases.py::test_unary[atanh(isfinite(real(x_i)) and real(x_i) != 0 and imag(x_i) is NaN) -> NaN + NaN j]
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +infinity and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and imag(x_i) is +infinity) -> ±0 + πj/2]
--- a/array-api-tests-macos-xfails.txt
+++ b/array-api-tests-macos-xfails.txt
@@ -1,3 +1,5 @@
+# Additional failures on MacOS: all tests listed in `array-api-tests-xfails.txt` fail everywhere, plus those listed here.
+#
 # https://github.com/numpy/numpy/issues/32234
 array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
 array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j]
--- a/array-api-tests-macos-xfails.txt
+++ b/array-api-tests-macos-xfails.txt
@@ -7,3 +7,4 @@ array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is 1 and imag(
 array_api_tests/test_special_cases.py::test_unary[atanh(isfinite(real(x_i)) and real(x_i) != 0 and imag(x_i) is NaN) -> NaN + NaN j]
 array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +infinity and imag(x_i) is NaN) -> +0 + NaN j]
 array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and imag(x_i) is +infinity) -> \xb10 + \u03c0j/2]
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and isfinite(imag(x_i))) -> NaN + NaN j]
--- a/array-api-tests-macos-xfails.txt
+++ b/array-api-tests-macos-xfails.txt
@@ -6,4 +6,4 @@ array_api_tests/test_operators_and_elementwise_functions.py::test_sqrt
 array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is 1 and imag(x_i) is +0) -> +infinity + 0j]
 array_api_tests/test_special_cases.py::test_unary[atanh(isfinite(real(x_i)) and real(x_i) != 0 and imag(x_i) is NaN) -> NaN + NaN j]
 array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +infinity and imag(x_i) is NaN) -> +0 + NaN j]
-array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and imag(x_i) is +infinity) -> ±0 + πj/2]
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and imag(x_i) is +infinity) -> \xb10 + \u03c0j/2]
--- a/array-api-tests-macos-xfails.txt
+++ b/array-api-tests-macos-xfails.txt
@@ -1,5 +1,5 @@
 # Additional failures on MacOS: all tests listed in `array-api-tests-xfails.txt` fail everywhere, plus those listed here.
-#
+
 # https://github.com/numpy/numpy/issues/32234
 array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
 array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j]
--- a/array-api-tests-windows-xfails.txt
+++ b/array-api-tests-windows-xfails.txt
@@ -1,3 +1,5 @@
+# Additional failures on Windows: all tests listed in `array-api-tests-xfails.txt` fail everywhere, plus those listed here.
+#
 # https://github.com/numpy/numpy/issues/32234
 array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
 array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j]
--- a/array-api-tests-windows-xfails.txt
+++ b/array-api-tests-windows-xfails.txt
@@ -1,0 +1,6 @@
+# https://github.com/numpy/numpy/issues/32234
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_creation_functions.py::test_arange
+array_api_tests/test_creation_functions.py::test_full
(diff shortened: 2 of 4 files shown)


array-api-tests-xfails.txt
```diff
@@ -51,3 +51,12 @@ array_api_tests/test_dlpack.py::test_dunder_dlpack
 
 # NumPy 1.26: TypeError: numpy.from_dlpack() takes no keyword arguments
 array_api_tests/test_dlpack.py::test_from_dlpack
+
+# Platform specific issues tracked in NumPy here: https://github.com/numpy/numpy/issues/32234
```
ev-br:
```suggestion
# https://github.com/numpy/numpy/issues/32234 : macOS specific failures
```

array-api-tests-xfails.txt
```diff
@@ -51,3 +51,12 @@ array_api_tests/test_dlpack.py::test_dunder_dlpack
...
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is 1 and imag(x_i) is +0) -> +infinity + 0j]
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +infinity and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and imag(x_i) is +infinity) -> \xb10 + \u03c0j/2]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_operators_and_elementwise_functions.py::test_sqrt
```
ev-br:
```suggestion

# https://github.com/numpy/numpy/issues/32234 : macOS specific failure
array_api_tests/test_operators_and_elementwise_functions.py::test_sqrt
```

ev-br (approved):

array-api-tests-xfails.txt
prady0t: I think some of them were also failing on Windows?

ev-br:
Ah, you're adding the new CI runs here... because I did not press the green button. OK, let's carry on in this PR then.

array-api-tests-xfails.txt
ev-br: Yeah OK, may add a comment as well. Or leave

prady0t:
Two new failures in CI

macOS:

```
array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and isfinite(imag(x_i))) -> NaN + NaN j] - AssertionError: out=(nan+1.5707963267948966j), but should be NaN + NaN j [atanh()]
```

Windows:

```
array_api_tests/test_creation_functions.py::test_arange - AssertionError: out.dtype=int64, should be default integer dtype int32 [arange()]

array_api_tests/test_creation_functions.py::test_full - AssertionError: out.dtype=int64, should be default integer dtype int32 [full()]
```

array-api-tests-xfails.txt
prady0t: Let's leave it as is to avoid confusion, as some fail on macOS and others on Windows. I thought the error log was similar for both OS.

prady0t:
On my machine (macOS 26.5.1) it passed

```
array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is NaN and isfinite(imag(x_i))) -> NaN + NaN j] PASSED
``` 
For test on windows:

`full((),0)` and `arange(0,0)` are problematic.

array-api-tests-xfails.txt
```diff
@@ -51,3 +51,14 @@ array_api_tests/test_dlpack.py::test_dunder_dlpack
 
 # NumPy 1.26: TypeError: numpy.from_dlpack() takes no keyword arguments
 array_api_tests/test_dlpack.py::test_from_dlpack
+
+# https://github.com/numpy/numpy/issues/32234 : macOS specific failures
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
```
betatim: For my education: how does the xfail machinery know that these tests should only be xfailed on macOS?

If possible, it would be good to not have a list of xfails that is "optional" (if it passes that is fine, if it fails that is also fine). In pytest there is a "strict" mode for xfails where something that is marked as failing but passes is a failure (🔴 ➡️ 🟢 ➡️ 🔴 - good luck not getting confused :D) . Maybe it is beyond the scope of this PR to get array-api-strict to run in strict mode, but we should aim for that ideal state. So we should briefly consider it here for OS dependent xfails.

prady0t: > For my education: how does the xfail machinery know that these tests should only be xfailed on macOS?

It doesn't; it loosely expects it to fall evrywhere. Yes, if we use `--strict`, it would raise an error if it passes.

Maybe we can modify test collection behaviour in the `contest.py` file? Or, use something like:

```py
@pytest.mark.skipif(sys.platform == "the platform")
```

On the tests that are failing specifically on certain platforms.

betatim: Yeah maybe [`pytest.mark.xfail`](https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) is the way to go here. They even use platform as the example for that decorator :D

ev-br: I'd have a rather strong desire to keep all xfails/skips in the same file. Tests themselves are in a different repository (array-api-tests), are reused across multiple array libraries, and themselves have zero `pytest.mark.{skip, xfail}` markers. All skipping/failing is done via these text files, and is specific to array libraries (compare to `-{xfail,skip}.txt` files in https://github.com/data-apis/array-api-compat).
Moreover, whether entries in a file are skips or xfails is controllable at the test suite invocation time by an environment variable [1].
IOW, whether to skip a test or not is a decision of whoever runs the test suite, when they run the test suite.

Therefore, if we are seriously looking to enable platform dependence, we need a DSL for the `.txt` files themselves. Currently, there isn't one, and the whole logic is here:
https://github.com/data-apis/array-api-tests/blob/master/conftest.py#L166-L200

[1] this is important from performance POV: with hypothesis, `xfails` are be very expensive, and needlessly so, cf https://github.com/data-apis/array-api-compat/pull/321

betatim: Ah, I didn't realise that the list of xfails is about tests from the `array-api-tests` repo, not tests that live in this repo.

How about extending the `--xfails-file` argument so that it can be passed multiple times? Then we can maintain a file that contains xfails that apply to all platforms and one each per platform that contains tests that only xfail on that platform. Something like `pytest .. --xfails-file general-xfails.txt --xfails-file macos-only-xfails.txt` ? I've not poked around pytest's arg handling enough to know if it support it. Otherwise we might need a recipe for combining two files into one on the fly. We could document this in this (array-api-strict) repo for those who want to invoke the array-api-tests tests on array-api-strict

betatim: I'm a hypothesis luddite, so I have nothing useful to say about the "use skips even if you mean xfail because xfails make things slow" :-/

ev-br: > Something like pytest .. --xfails-file general-xfails.txt --xfails-file macos-only-xfails.txt ?

Oh this is a neat idea! Definitely much better than inventing a DSL for a single `xfails.txt` file and parsing it out the comments or something.

To test-drive the idea, there are `--skips-file=...` and  `--xfails-file=...` already, so they can be reused :-).
This would be a hack, of course, so longer term I agree it'd be better to be able to pass a semantically correct file multiple times. I haven't try to see how pytest handles these, but it should be very much doable I think. 

> I'm a hypothesis luddite, so I have nothing useful to say about the "use skips even if you mean xfail because xfails make things slow" :-/

Well, in the current setup the user keeps using semantically meaningful `--xfails-file` name. There's an additional switch to toggle the internal details of how pytest and hypothesis handle an expected failure. This switch is obscure enough (an environment variable) to be only used deliberately, when pressed hard enough. An "anecdotal evidence" from https://github.com/data-apis/array-api-tests#turning-xfails-into-skips is real: we did manage to increase the number of examples by a factor of 4-5 and keep a reasonable tun time for some backends. And with hypothesis the number of examples is the measure of reliability: at some point Dask CI had 10 examples, and that meant it was effectively untested.

betatim: I put AI to work to find an answer to the "can we use multiple files" question (after many many tokens burnt and one espresso for me): "Short answer: yes, it works, but only if the option is re-registered with `action="append"` — which is a two-line change in `array-api-tests/conftest.py`."

Should I make a PR?

ev-br: Emphatically yes!

ev-br: Implemented in https://github.com/data-apis/array-api-tests/pull/454, resolving.

ev-br (changes requested):
Now that https://github.com/data-apis/array-api-tests/pull/454 implemented support for multiple skip files, would be best to move platform-specific xfails to the new `...-macos-xfail.txt` files.

prady0t:
The CI is green now, should be good to go!

array-api-tests-windows-xfails.txt
```diff
@@ -0,0 +1,6 @@
+# https://github.com/numpy/numpy/issues/32234
+array_api_tests/test_special_cases.py::test_unary[atanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is +infinity) -> +0 + NaN j]
+array_api_tests/test_special_cases.py::test_unary[tanh(real(x_i) is +0 and imag(x_i) is NaN) -> +0 + NaN j]
+array_api_tests/test_creation_functions.py::test_arange
```
ev-br, line 9: This and the one below are a bit different from unary/binary functions at magic values of the arguments. Judging by your comment, https://github.com/data-apis/array-api-strict/pull/240#issuecomment-5246873145, these two smoke out some "empty array" fast paths that deviate from the general position. Ok to skip here for now, but could you please investigate whether these come from -strict or numpy? If the latter, and visible in the numpy main, could you report them to numpy, too? If it's -strict, let's fix them here.

prady0t: Looks like these two only fail on a specific Python version (3.10).

prady0t: Correction: Fails with numpy version 1.26 on Windows (v1.26 only ever runs with Python 3.10 in CI)

prady0t: The reason is that NumPy on Windows chooses `int32` as the default dtype, whereas the test suite expects int64 as the default.
This was fixed in NumPy 2.0: https://github.com/numpy/numpy/pull/24224

How should we proceed? Remove skips and CI, or keep it in the xfail file with a comment?

ev-br: Thanks for checking it!
I think it's best to keep them xfailed on windows with a comment that both were fixed in numpy 2.0.
Then we'll have enough info to remove them when we stop supporting numpy 1.26 (no definite plans ATM, so "some day").

Once that's done, let's merge this PR.

array-api-tests-macos-xfails.txt
```diff
@@ -0,0 +1,10 @@
+# https://github.com/numpy/numpy/issues/32234
```
ev-br, line 3:
```suggestion
# Additional failures on MacOS: all tests listed in `array-api-tests-xfails.txt` fail everywhere, plus those listed here.
#
# https://github.com/numpy/numpy/issues/32234
```

array-api-tests-windows-xfails.txt
```diff
@@ -0,0 +1,8 @@
+# https://github.com/numpy/numpy/issues/32234
```
ev-br, line 3:
```suggestion
# Additional failures on Windows: all tests listed in `array-api-tests-xfails.txt` fail everywhere, plus those listed here.
#
# https://github.com/numpy/numpy/issues/32234
```

ev-br (approved):
LGTM plus a small suggestion below (feel free to edit the wording if you prefer).

prady0t:
And it's done.

ev-br (approved):

ev-br:
And it's merged, thank you @prady0t

